PR #28, opened 2022-04-06
Add workflow steps to build bundle and catalog

fabiendupont:
When the operator image is updated, the bundle image needs to be updated
as well to reflect the new operator image digest. This change adds this
by calling `make bundle` and rebuilding / pushing the bundle image.

On the same line, the catalog image needs to be rebuilt to reflect the
new bundle image digest. This is also part of this change.

The `bundle` folder is also ignored since its content is generated and  
uses the operator image digest, that changes with every merge.

There can be additional improvements to ensure digests are used
everywhere, but it's a first stab at automating the operator > bundle >
catalog flow.

Signed-off-by: Fabien Dupont <fabiendupont@pm.me>

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -14,11 +14,35 @@ jobs:
       packages: write
 
     steps:
-      - name: Generate image metadata
-        id: meta
+      - name: Generate image metadata for operator
+        id: meta-operator
         uses: docker/metadata-action@v3
         with:
           images: ghcr.io/qbarrand/oot-operator
+          tags: |
+            type=ref,event=branch
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+            type=sha
+
+      - name: Generate image metadata for bundle
+        id: meta-bundle
+        uses: docker/metadata-action@v3
+        with:
+          images: ghcr.io/qbarrand/oot-operator-bundle
+          tags: |
+            type=ref,event=branch
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+            type=sha
+
+      - name: Generate image metadata for catalog
+        id: meta-catalog
+        uses: docker/metadata-action@v3
+        with:
+          images: ghcr.io/qbarrand/oot-operator-catalog
           tags: |
             type=ref,event=branch
             type=semver,pattern={{version}}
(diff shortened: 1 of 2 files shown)


.github/workflows/container-image.yml
```diff
@@ -36,10 +64,48 @@ jobs:
...
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.meta-operator.outputs.tags }}
+          labels: ${{ steps.meta-operator.outputs.labels }}
+
+      - name: Setup Go ${{ env.GOLANG_VERSION }}
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GOLANG_VERSION }}
```
qbarrand, line 79: Not sure we need an environment variable here.
```suggestion
      - name: Setup Go
        uses: actions/setup-go@v3
        with:
          go-version: 1.18
```

qbarrand (commented):
Thank you!

.github/workflows/container-image.yml
```diff
@@ -36,10 +64,48 @@ jobs:
...
+          export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/${OPERATOR_SDK_VERSION}
+          curl -LO ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH}
+          mkdir -p ${HOME}/.local/bin
+          chmod +x operator-sdk_${OS}_${ARCH} && sudo mv operator-sdk_${OS}_${ARCH} ${HOME}/.local/bin/operator-sdk
+          echo "${HOME}/.local/bin" >> $GITHUB_PATH
+
+      - name: Generate the bundle content
+        run: make bundle
+        env:
+          IMG: ghcr.io/qbarrand/oot-operator@${{ steps.build-push-operator.outputs.digest }}
```
qbarrand, line 95: I would in fact do that outside this workflow and just compare if that command generated a diff on the bundle files.
The git repo should always be the source of truth.

.github/workflows/container-image.yml
fabiendupont: You won't have Go 1.18 in RHEL before RHEL 8.7, i.e. Nov 2022.

.github/workflows/container-image.yml
fabiendupont: It will generate a diff because the `IMG` variable is linked to the new operator image.
The `bundle` folder is generate from the kustomize files, so the source of truth is the `config` folder.

.github/workflows/container-image.yml
```diff
@@ -10,12 +10,16 @@ jobs:
     name: Build and push
     runs-on: ubuntu-latest
 
+    env:
+      GOLANG_VERSION: 1.17
+      VERSION: ${{ needs.get-operator-version.outputs.version }}
```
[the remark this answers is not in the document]
qbarrand: I was rather thinking about `VERSION`. It depends on a job that is not defined in the workflow.

.github/workflows/container-image.yml
fabiendupont: Just to have the same version in other steps if we need it. I agree it's a stretch.

.github/workflows/container-image.yml
fabiendupont: Yes. I've added it to `.gitignore`.

PR #29, opened 2022-04-06
Move the Module API to v1alpha1

qbarrand:
This PR moves the `Module` API to `v1alpha1` as it is still under development.


mresvanis:
Very nice, I just tested this branch on my dev cluster along with the GPU Operator v2 and it works like a charm :+1:

PR #42, opened 2022-04-12
Add support for build secrets

### LINKED ISSUE #38: In-cluster builds: add build secrets
Kaniko builds are able to read secrets in the pod's `/run/secrets` directory. We could mount secrets there to be used during the build without being in the final image.

qbarrand:
Mount build secrets inside `/run/secrets` in the builder pod.

Closes #38.


api/v1alpha1/module_types.go
```diff
@@ -58,6 +58,10 @@ type Build struct {
 
 	// Push contains settings determining how to push a built DriverContainer image.
 	Push PushOptions `json:"push"`
+
+	// Secrets is an optional list of secrets to be made available to the build system.
+	// +optional
+	Secrets []v1.LocalObjectReference `json:"secrets"`
```
fabiendupont, line 64: This limits the object reference to the same namespace, but what does it mean for non namespaced resources? I think the entity creating the `Module` CR will own the secret and store them in its own namespace.

qbarrand: Created #43 to transition `Module` to a namespaced resource.

PR #69, opened 2022-05-30
MGMT-10532 generate mocks for client interface

erusso7:
cc: @ybettan


controllers/client.go
```diff
@@ -0,0 +1,9 @@
+package controllers
+
+import "sigs.k8s.io/controller-runtime/pkg/client"
+
+//go:generate mockgen -source=client.go -package=controllers -destination=mock_client.go
+
+type Client interface {
+	client.Client
+}
```
erusso7, line 9: This is the only way I managed to generate *only* the mock for the `client.Client` interface because if we use the original source file `sigs.k8s.io/controller-runtime@v0.11.2/pkg/client/interfaces.go` it generates the code for 8 interfaces.

controllers/daemonset_test.go
```diff
@@ -275,24 +280,15 @@ var _ = Describe("daemonSetGenerator", func() {
...
-			afterDeleteDaemonSetList := appsv1.DaemonSetList{}
-
-			Expect(
-				client.List(context.TODO(), &afterDeleteDaemonSetList),
-			).To(
-				Succeed(),
-			)
-
-			Expect(afterDeleteDaemonSetList.Items).To(HaveLen(1))
-			Expect(afterDeleteDaemonSetList.Items[0].Name).To(Equal(legitName))
```
erusso7, line 288: This piece of code is testing the implementation of the `client.Client`, in this case, it's the `fakeClient` and we don't want to test the client behavior, instead, we'll only rely on its return.

ybettan:
Please make sure you squash all your commit and keep a clean commit - see https://github.com/openshift/special-resource-operator/blob/master/CONTRIBUTING.md

controllers/client.go
ybettan: Just to make sure I understand, how do you generate the mock by using `sigs.k8s.io/controller-runtime@v0.11.2/pkg/client/interfaces.go`?

Also I would suggest moving this file to `pkg/client/client.go` but as all our code is currently in `controller/` for some reason maybe it is better to leave it here and have another task that moves all the packages to `pkg/` across the whole code.

Toughts @yevgeny-shnaidman?

controllers/daemonset_test.go
```diff
@@ -259,7 +263,8 @@ var _ = Describe("daemonSetGenerator", func() {
 				ObjectMeta: metav1.ObjectMeta{Name: notLegitName, Namespace: namespace},
 			}
 
-			client := fake.NewClientBuilder().WithObjects(&dsLegit, &dsNotLegit).Build()
+			client := controllers.NewMockClient(gomockCtrl)
+			client.EXPECT().Delete(context.TODO(), gomock.Any()).Return(nil)
```
ybettan, line 267:
```suggestion
			client.EXPECT().Delete(context.TODO(), existingDS[1]).Return(nil)
```

controllers/daemonset_test.go
```diff
@@ -275,24 +280,15 @@ var _ = Describe("daemonSetGenerator", func() {
 			).To(
```
ybettan, line 280: I will suggest to also check that `GarbageCollect` didn't return any error.

controllers/client.go
yevgeny-shnaidman: I think we can make all the changes in one PR, lets just rename the PR and Jira accordingly

controllers/daemonset_test.go
erusso7: I'm not a big fan of doing it because we are coupling the tests to the implementation of the code. 
The idea behind this shouldn't be "I want you to call a specific method with specific args", instead should be "I don't care what you do, I want you to return what DS was deleted".

I don't know if I explain myself properly :joy:

controllers/client.go
erusso7: done :heavy_check_mark:

controllers/build/job/manager_test.go
```diff
@@ -177,22 +187,6 @@ var _ = Describe("JobManager", func() {
...
-				client.Get(ctx, key, &newJob),
-			).To(
-				Succeed(),
-			)
-
-			Expect(
-				cmp.Diff(j, newJob),
-			).To(
-				BeEmpty(),
-			)
```
erusso7, line 195: These assertions have blown my mind. If I'm not wrong, it's checking what the `fake.Client` would return if we call the `Get()` method after creating the job. 
If that assumption is correct, I would say that's not the way to go. As a unit test, it should only rely on the returned result instead of the internal state of the dependencies.
WDYT?

yevgeny-shnaidman: before your change, the unit test was based on the faked client, which is an actual client backed by a very simple object store. So this assertions were actually check what was actually stored in the object store. since we are moving to the real mocks, these assertions become obsolete

controllers/node_kernel_reconciler_test.go
```diff
@@ -40,24 +51,26 @@ var _ = Describe("NodeKernelReconciler", func() {
...
 			nkr := controllers.NewNodeKernelReconciler(client, labelName)
 			req := ctrl.Request{
 				NamespacedName: types.NamespacedName{Name: nodeName},
 			}
 
-			ctx := context.TODO()
-
 			res, err := nkr.Reconcile(ctx, req)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(res).To(Equal(res))
```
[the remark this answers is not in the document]
yevgeny-shnaidman: quite possible is it just a big in the unit test code :)

controllers/node_kernel_reconciler_test.go
```diff
@@ -40,24 +51,26 @@ var _ = Describe("NodeKernelReconciler", func() {
...
-
-			updatedNode := v1.Node{}
-
-			err = client.Get(ctx, types.NamespacedName{Name: nodeName}, &updatedNode)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(updatedNode.Labels).To(HaveKeyWithValue(labelName, kernelVersion))
```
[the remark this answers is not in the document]
ybettan: Agree to remove it and only test the input/output if the function which is being tested.

controllers/module_reconciler_test.go
```diff
@@ -276,16 +275,28 @@ var _ = Describe("ModuleReconciler", func() {
...
+					},
+				),
+				client.EXPECT().List(ctx, gomock.Any(), gomock.Any()).DoAndReturn(
+					func(c context.Context, list *v1.NodeList, o interface{}) error {
+						list.Items = nodeList.Items
+						return nil
+					},
+				),
+				client.EXPECT().Get(ctx, gomock.Any(), gomock.Any()).Return(nil),
+				client.EXPECT().Patch(ctx, gomock.Any(), gomock.Any()).Return(nil),
```
erusso7, line 296: These two method calls were the most tricky to find because they are happening in the implementation of a hardcoded external dependency, instead of using an interface. The `Get()` and `Patch()` methods are used in the implementation of the 
`controllerutil.CreateOrPatch()` method.

controllers/node_module_mapper_test.go
```diff
@@ -67,11 +89,17 @@ var _ = Describe("FindModulesForNode", func() {
 			},
 		}
 
-		nmm := controllers.NewNodeModuleMapper(
-			fake.NewClientBuilder().WithScheme(scheme).WithObjects(&mod1, &mod2, &node).Build(),
-			logr.Discard(),
+		client := client.NewMockClient(ctrl)
+		client.EXPECT().List(context.TODO(), gomock.Any(), gomock.Any()).DoAndReturn(
+			func(ctx context.Context, list *v1alpha1.ModuleList, _ interface{}) error {
```
erusso7, line 94: This `func` MUST have 3 arguments, otherwise the testing library, by reflection, will raise this error:
```
Test Panicked
  In [It] at: /usr/local/go/src/reflect/value.go:398
  reflect: Call with too few input arguments
```

So, In [the implementation](https://github.com/qbarrand/oot-operator/pull/69/commits/9983d58efefa17c8b8d350283c266144228a7c44#diff-2e7facd3081a7f636bc0a916dea254e506b6d27488e7e29d1046f23037c0ae27R36) I needed to add a third parameter `nil` to satisfy this requirement.

It's not nice when the test forces us to code something unnecessary. Thoughts on this? 
@ybettan @yevgeny-shnaidman

ybettan:
Please follow https://github.com/qbarrand/oot-operator/blob/main/CONTRIBUTING.md for the commit message format.

ybettan:
@yevgeny-shnaidman Should we remove `pkg/test/client.go` as part of this PR?

pkg/client/client.go
```diff
@@ -0,0 +1,9 @@
+package client
+
+import k8sClient "sigs.k8s.io/controller-runtime/pkg/client"
+
+//go:generate mockgen -source=client.go -package=client -destination=mock_client.go
+
+type Client interface {
```
qbarrand, line 7: Not sure this type is needed - `gomock` can directly generate a mock for `sigs.k8s.io/controller-runtime/pkg/client.Client`.

yevgeny-shnaidman: i think that it is a good practice to create a Client layer for accessing the cluster: that way we can use whatever client we need (besides controller-runtime), without changing the rest of the code. As we saw in SRO , we needed to use more then one client implementation

qbarrand: Actually, I don't see any benefit in using several client implementations - controller-runtime's is generic enough and works with all object types.

ybettan: > gomock can directly generate a mock for sigs.k8s.io/controller-runtime/pkg/client.Client

@qbarrand How do you do that without adding the mockgen annotation to `sigs.k8s.io/controller-runtime/pkg/client.Client`?

qbarrand: @ybettan in any file:
```
//go:generate mockgen -destination mock_client.go sigs.k8s.io/controller-runtime/pkg/client Client
```

erusso7: The idea of this internal type was because I didn't know how to specify only one interface and the `mockgen` was generating mocks for 7 different interfaces. Now, as @qbarrand indicated, we can specify only the needed interface `Client`, so I would go with that approach instead of creating an internal type :+1:

ybettan:
`e2e / In-cluster build (pull_request)` is flaky. You can just rerun it when you see it fails unrelatedly to your changes.

ybettan:
/hold
Don't forget to squash the commits.

internal/client/client.go
```diff
@@ -0,0 +1,3 @@
+package client
+
+//go:generate mockgen -package=client -destination mock_client.go sigs.k8s.io/controller-runtime/pkg/client Client
```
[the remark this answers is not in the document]
qbarrand: Will `go generate` see it if it is in a test file?

module-ooto-ci.yaml
```diff
@@ -0,0 +1,25 @@
+---
```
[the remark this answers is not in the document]
qbarrand: Indeed, I think this file should be removed.

controllers/module_reconciler_test.go
```diff
@@ -64,29 +64,32 @@ var _ = Describe("ModuleReconciler", func() {
...
+					},
+				),
+				client.EXPECT().List(ctx, gomock.Any(), gomock.Any()).DoAndReturn(
+					func(_ interface{}, list *ootov1alpha1.ModuleList, _ ...interface{}) error {
+						list.Items = []ootov1alpha1.Module{mod}
+						return nil
+					},
+				),
+				mockMetrics.EXPECT().SetExistingKMMOModules(1),
+				client.EXPECT().List(ctx, gomock.Any(), gomock.Any()).Return(nil),
```
qbarrand, line 85:
```suggestion
				client.EXPECT().List(ctx, gomock.Any(), gomock.Any()),
```

controllers/module_reconciler_test.go
```diff
@@ -116,20 +119,31 @@ var _ = Describe("ModuleReconciler", func() {
...
```
qbarrand, line 139:
```suggestion
				client.EXPECT().List(ctx, gomock.Any(), gomock.Any()),
```

controllers/module_reconciler_test.go
```diff
@@ -179,16 +193,33 @@ var _ = Describe("ModuleReconciler", func() {
...
+					},
+				),
+				mockMetrics.EXPECT().SetExistingKMMOModules(0),
+				client.EXPECT().List(ctx, gomock.Any(), gomock.Any()).DoAndReturn(
+					func(_ interface{}, list *v1.NodeList, _ ...interface{}) error {
+						list.Items = nodeList.Items
+						return nil
+					},
+				),
+				client.EXPECT().Get(ctx, gomock.Any(), gomock.Any()).Return(nil),
```
qbarrand, line 219:
```suggestion
				client.EXPECT().Get(ctx, gomock.Any(), gomock.Any()),
```

erusso7:
> /hold Don't forget to squash the commits.

Yes, I was to test the `e2e / Prebuilt kernel module (NFD: false)` test :)

controllers/module_reconciler_test.go
```diff
@@ -272,21 +296,39 @@ var _ = Describe("ModuleReconciler", func() {
...
+				),
+				mockMetrics.EXPECT().SetExistingKMMOModules(1),
+				client.EXPECT().List(ctx, gomock.Any(), gomock.Any()).DoAndReturn(
+					func(_ interface{}, list *v1.NodeList, _ ...interface{}) error {
+						list.Items = nodeList.Items
+						return nil
+					},
+				),
+				client.EXPECT().Get(ctx, gomock.Any(), gomock.Any()).Return(nil),
+				client.EXPECT().Patch(ctx, gomock.Any(), gomock.Any()).Return(nil),
```
qbarrand, line 324:
```suggestion
				client.EXPECT().Get(ctx, gomock.Any(), gomock.Any()),
				client.EXPECT().Patch(ctx, gomock.Any(), gomock.Any()),
```

module-ooto-ci.yaml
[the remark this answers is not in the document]
erusso7: Exactly, I thought the same. I think it's something that I carried from the `rebase` because I don't remember modifying this file :thinking: 
Should I remove it?

controllers/node_kernel_reconciler_test.go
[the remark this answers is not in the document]
qbarrand: Created #88 as a follow-up for this.

controllers/node_kernel_reconciler_test.go
```diff
@@ -69,24 +82,27 @@ var _ = Describe("NodeKernelReconciler", func() {
...
+			client := client.NewMockClient(gCtrl)
+			gomock.InOrder(
+				client.EXPECT().Get(ctx, gomock.Any(), gomock.Any()).DoAndReturn(
+					func(_ interface{}, _ interface{}, n *v1.Node) error {
+						n.ObjectMeta = node.ObjectMeta
+						n.Status = node.Status
+						return nil
+					},
+				),
+				client.EXPECT().Patch(ctx, gomock.Any(), gomock.Any()).Return(nil),
```
qbarrand, line 95:
```suggestion
				client.EXPECT().Patch(ctx, gomock.Any(), gomock.Any()),
```

controllers/node_module_mapper_test.go
```diff
@@ -0,0 +1,100 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var _ = Describe("FindModulesForNode", func() {
+	var ctrl *gomock.Controller
+	BeforeEach(func() {
+		ctrl = gomock.NewController(GinkgoT())
```
qbarrand, line 23: You may want to declare `client` here and initialize it in `BeforeEach` as well.

controllers/node_module_mapper_test.go
```diff
@@ -0,0 +1,100 @@
...
+	})
+
+	It("should return nothing if there are no modules", func() {
+		client := client.NewMockClient(ctrl)
+		client.EXPECT().List(context.TODO(), gomock.Any(), gomock.Any()).Return(nil)
```
qbarrand, line 28:
```suggestion
		client.EXPECT().List(context.TODO(), gomock.Any(), gomock.Any())
```

internal/build/job/manager_test.go
```diff
@@ -145,8 +148,10 @@ var _ = Describe("JobManager", func() {
 				getter.EXPECT().ImageExists(ctx, imageName, po),
 				maker.EXPECT().MakeJob(mod, km.Build, kernelVersion, km.ContainerImage).Return(nil, errors.New("random error")),
 			)
+			client := client.NewMockClient(ctrl)
+			client.EXPECT().List(ctx, gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
```
qbarrand, line 152:
```suggestion
			client.EXPECT().List(ctx, gomock.Any(), gomock.Any(), gomock.Any())
```

internal/build/job/manager_test.go
```diff
@@ -177,7 +182,11 @@ var _ = Describe("JobManager", func() {
 				maker.EXPECT().MakeJob(mod, km.Build, kernelVersion, km.ContainerImage).Return(&j, nil),
 			)
 
-			client := fake.NewClientBuilder().Build()
+			client := client.NewMockClient(ctrl)
+			gomock.InOrder(
+				client.EXPECT().List(ctx, gomock.Any(), gomock.Any(), gomock.Any()).Return(nil),
+				client.EXPECT().Create(ctx, &j).Return(nil),
```
qbarrand, line 188:
```suggestion
				client.EXPECT().List(ctx, gomock.Any(), gomock.Any(), gomock.Any()),
				client.EXPECT().Create(ctx, &j),
```

internal/daemonset/daemonset_test.go
```diff
@@ -2,18 +2,23 @@ package daemonset
...
 
 	"github.com/go-openapi/swag"
+
+	"github.com/golang/mock/gomock"
 	"github.com/google/go-cmp/cmp"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	ootov1alpha1 "github.com/qbarrand/oot-operator/api/v1alpha1"
+
+	// "github.com/qbarrand/oot-operator/controllers"
```
qbarrand, line 15:
```suggestion
```

module-ooto-ci.yaml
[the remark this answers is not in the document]
qbarrand: Yes please.

qbarrand:
I think all `.Return(nil)` in mocked calls can be removed, that's the default behaviour. I did not mark them all in the review.

internal/daemonset/daemonset_test.go
```diff
@@ -24,9 +28,16 @@ const (
 	devicePluginImage = "device-plugin-image"
 )
 
+var ctrl *gomock.Controller
+
 var _ = Describe("SetDriverContainerAsDesired", func() {
+	var gomockCtrl *gomock.Controller
```
qbarrand, line 34: Why do we have two `*gomock.Controller` here?

erusso7: not anymore

internal/client/client.go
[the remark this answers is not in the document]
qbarrand: I'm rather happy with having a simple file with that comment. Maybe putting it into an `internal/testutil/client` package would be a bit more explicit though.

internal/daemonset/daemonset_test.go
```diff
@@ -443,11 +584,16 @@ var _ = Describe("GarbageCollect", func() {
...
+		clnt = client.NewMockClient(ctrl)
+	})
+
 	It("should return an empty map if no DaemonSets are present", func() {
-		dc := NewCreator(
-			fake.NewClientBuilder().WithScheme(scheme).Build(),
-			kernelLabel,
-			scheme)
+		opts := []interface{}{gomock.Any(), gomock.Any()}
+		clnt.EXPECT().List(context.TODO(), gomock.Any(), opts...)
```
qbarrand, line 594: Same here, do we need to declare `opts`?

internal/client/client.go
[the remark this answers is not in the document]
erusso7: After discussing it offline, we'll leave it as it is for the moment. Because it's an `internal` file it shouldn't be imported. We might re-consider this in the future.

qbarrand (approved):

controllers/module_reconciler_test.go
```diff
@@ -28,6 +28,7 @@ var _ = Describe("ModuleReconciler", func() {
 	Describe("Reconcile", func() {
 		var (
 			ctrl        *gomock.Controller
+			clnt        *client.MockClient
```
ybettan, line 31:
```suggestion
			client        *client.MockClient
```

ybettan (approved):

PR #70, opened 2022-05-31
MGMT-10518: separate Driver container and Device Plugin into 2 daemon…

yevgeny-shnaidman:
…sets

This PR contains the following:
1) separate driver container and device plugin container into 2
   daemonsets
2) refactor module reconciliation code that deals with drvier container,
   device plugin and build
3) refactor daemonset code that is responsible for creating daemonset
   structure for driver container or device plugin
4) add unit test


ybettan:
Are you OK with sticking to that [format](https://github.com/openshift/special-resource-operator/blob/master/CONTRIBUTING.md) for commit messages?

In the u/s repo (more than in the d/s repo) Jira tickets have less visibility and we expect contributor also out of Redhat to create PRs.

WDYT?

yevgeny-shnaidman:
> Are you OK with sticking to that [format](https://github.com/openshift/special-resource-operator/blob/master/CONTRIBUTING.md) for commit messages?
> 
> In the u/s repo (more than in the d/s repo) Jira tickets have less visibility and we expect contributor also out of Redhat to create PRs.
> 
> WDYT?

For now i prefer to stick to this format, it helps to quickly identify the ticket, especially if the commit message
does not provide detailed explanations

controllers/daemonset.go
```diff
@@ -171,44 +154,91 @@ func (dc *daemonSetGenerator) SetAsDesired(ds *appsv1.DaemonSet, image string, m
...
+	}
+
+	return dc.constructDaemonSet(ctx, ds, mod, *mod.Spec.DevicePlugin, "device-plugin", "", standardLabels,
+		mod.Spec.Selector, containerVolumeMounts, dsVolumes, true)
+}
+
+func (dc *daemonSetGenerator) constructDaemonSet(ctx context.Context,
+	ds *appsv1.DaemonSet,
+	mod *ootov1alpha1.Module,
+	container v1.Container,
```
ybettan, line 201: Do we really need it? It is already part of the module object passed to that function.

yevgeny-shnaidman: this is a common function, used both for creating DevicePlugin daemonset and DrvierContainer daemonset. If we don't pass the container, then we need to pass a flag, that will tell the code which container to use.  So i prefer to pass the container directly, instead of passing the flag and adding if clause into the code

controllers/module_reconciler.go
```diff
@@ -126,71 +180,79 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	kernelVersion string) (bool, error) {
+	if km.Build == nil {
+		return false, nil
 	}
 
-	//// TODO qbarrand: find a better place for this
-	//if err := r.su.SetAsReady(ctx, &mod, "TODO", "TODO"); err != nil {
-	//	return res, fmt.Errorf("could not set the initial conditions: %v", err)
-	//}
+	// TODO check access to the image - execute build only if needed
```
ybettan, line 194: Are we doing it in another PR?

yevgeny-shnaidman: yes.

ybettan (approved):

PR #73, opened 2022-06-08
Adding support for variable substituation in kernelmapping fields

yevgeny-shnaidman:
This PR contains the following:
1) generating kernel data per node(full version, major/minor/patch numbers,
   RHEL release versions, architecture) based on values stored by the kubelet
   in the node Status field
2) substiting the template variables defined in the modules kernel mappings
   with the appropriate values collected in step 1
3) unit tests

[MGMT-10537]

### DIFF
--- a/controllers/module/kernelmapper.go
+++ b/controllers/module/kernelmapper.go
@@ -1,17 +1,41 @@
 package module
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
 	"regexp"
+	"strings"
 
+	"github.com/a8m/envsubst/parse"
 	ootov1alpha1 "github.com/qbarrand/oot-operator/api/v1alpha1"
+	v1 "k8s.io/api/core/v1"
 )
+
+const (
+	kernelVersionMajorIdx = 0
+	kernelVersionMinorIdx = 1
+	kernelVersionPatchIdx = 2
+)
+
+type NodeOSConfig struct {
+	kernelFullVersion  string `subst:"KERNEL_FULL_VERSION"`
+	kernelVersionMMP   string `subst:"KERNEL_XYZ"`
+	kernelVersionMajor string `subst:"KERNEL_X"`
+	kernelVersionMinor string `subst:"KERNEL_Y"`
+	kernelVersionPatch string `subst:"KERNEL_Z"`
+	kernelRHELVersion  string `subst:"KERNEL_RHEL_VERSION"`
+	kernelRHELRelease  string `subst:"KERNEL_RHEL_RELEASE"`
+	kernelRHELArch     string `subst:"KERNEL_RHEL_ARCH"`
+}
 
 //go:generate mockgen -source=kernelmapper.go -package=module -destination=mock_kernelmapper.go
 
 type KernelMapper interface {
 	FindMappingForKernel(mappings []ootov1alpha1.KernelMapping, kernelVersion string) (*ootov1alpha1.KernelMapping, error)
+	GetNodeOSConfig(node *v1.Node) *NodeOSConfig
+	PrepareKernelMapping(mapping *ootov1alpha1.KernelMapping, osConfig *NodeOSConfig) (*ootov1alpha1.KernelMapping, error)
 }
 
 type kernelMapper struct{}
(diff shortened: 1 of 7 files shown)


ybettan:
`sed /ofr/for/` in the commit message (and PR title).

controllers/module_reconciler.go
```diff
@@ -173,6 +174,12 @@ func (r *ModuleReconciler) getKernelMappings(ctx context.Context, nodes v1.NodeL
 			continue
 		}
 
+		m, err = r.km.PrepareKernelMapping(m, osConfig)
+		if err != nil {
+			nodeLogger.Info("failed to substitute the tempate variable in the mapping", "error", err)
```
ybettan, line 179:
```suggestion
			nodeLogger.Info("failed to substitute the template variable in the mapping", "error", err)
```

controllers/module/kernelmapper.go
```diff
@@ -1,17 +1,41 @@
...
+	kernelRHELVersion  string `subst:"KERNEL_RHEL_VERSION"`
+	kernelRHELRelease  string `subst:"KERNEL_RHEL_RELEASE"`
+	kernelRHELArch     string `subst:"KERNEL_RHEL_ARCH"`
+}
+
 //go:generate mockgen -source=kernelmapper.go -package=module -destination=mock_kernelmapper.go
 
 type KernelMapper interface {
 	FindMappingForKernel(mappings []ootov1alpha1.KernelMapping, kernelVersion string) (*ootov1alpha1.KernelMapping, error)
+	GetNodeOSConfig(node *v1.Node) *NodeOSConfig
```
ybettan, line 37: Does it really need to be an interface method? The only one using its return value is `PrepareKernelMapping`.
Wouldn't it be better to make it a private method of the `KernelMapper` object instead?

controllers/module/kernelmapper.go
```diff
@@ -41,3 +65,50 @@ func (kernelMapper) FindMappingForKernel(mappings []ootov1alpha1.KernelMapping,
...
+
+	osConfigFieldsList := regexp.MustCompile("[.,-]").Split(node.Status.NodeInfo.KernelVersion, -1)
+
+	osConfig.fullKernelVersion = node.Status.NodeInfo.KernelVersion
+	osConfig.kernelVersionMMP = strings.Join(osConfigFieldsList[:kernelVersionPatchIdx+1], ".")
+	osConfig.kernelVersionMajor = osConfigFieldsList[kernelVersionMajorIdx]
+	osConfig.kernelVersionMinor = osConfigFieldsList[kernelVersionMinorIdx]
+	osConfig.kernelVersionPatch = osConfigFieldsList[kernelVersionPatchIdx]
+
+	if strings.Contains(node.Status.NodeInfo.OSImage, "Red Hat") {
```
ybettan, line 80: Is there another k8s distro that uses RHEL/RHCOS? If not, for OCP, won't client use the d/s operator instead? Do we really need to support Redhat u/s? Feels like a duplication effort with d/s don't you think?

controllers/module_reconciler.go
yevgeny-shnaidman: fixed

controllers/module/kernelmapper.go
yevgeny-shnaidman: We will probably need the output when we working on MGMT-10539, and probably when I will start looking into CONTAINER_IMAGE template ( currently , not sure if it is needed or not). That's why i made it public, in order for it to be called only once

controllers/module/kernelmapper.go
yevgeny-shnaidman: we would eventually also support Ubuntu and and other distributions. and currently i am treating this code as upstream  code. Once we have other distribution to test , we will expand this code

ybettan: I don't think we should expect u/s to be run on OCP. I think we should encourage clients to use the d/s project if they are using OCP. 

I am not aware of any other k8s distro (rather than OCP) that runs RHEL on the nodes so this is just misleading as we don't intend to keep support for OCP on this project (regarding all the security constrains etc).

@bthurber WDYT?

yevgeny-shnaidman: As agreed, add a TODO that we will remove that code from upstream once we have it in downstream

controllers/module/kernelmapper.go
```diff
@@ -41,3 +65,50 @@ func (kernelMapper) FindMappingForKernel(mappings []ootov1alpha1.KernelMapping,
...
+	err = json.Unmarshal([]byte(parsedMapping), &substMapping)
+	return &substMapping, err
+}
+
+func (k *kernelMapper) prepareOSConfigList(osConfig NodeOSConfig) []string {
+	t := reflect.TypeOf(osConfig)
+	v := reflect.ValueOf(osConfig)
+
+	varList := make([]string, t.NumField())
+	for i := 0; i < t.NumField(); i++ {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: actually using for loop for integer case is the more idiomatic way in Go. See https://groups.google.com/g/golang-nuts/c/7J8FY07dkW0/m/goWaNVOkQU0J

ybettan (approved):

PR #75, opened 2022-06-09
Adding support for pull/push secrets.

ybettan:
The operator may need to pull container images or to push some in case
of making use of the signing feature and then pushing the signed images.
In Both those cases we need the pull/push secret of the image registry.


ybettan:
Waiting on:
* https://github.com/qbarrand/oot-operator/pull/82 by @yevgeny-shnaidman 
* CRD update by @qbarrand

api/v1alpha1/module_types.go
```diff
@@ -125,6 +125,11 @@ type ModuleSpec struct {
 	// More info: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
 	// +optional
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+
+	// +optional
+	// ServiceAccountName is the name of the ServiceAccount to use to run this pod.
```
qbarrand, line 130: The comment does not match the field's name.

internal/build/job/maker.go
```diff
@@ -96,11 +96,11 @@ func (m *maker) MakeJob(mod ootov1alpha1.Module, buildConfig *ootov1alpha1.Build
 							Args:         args,
 							Name:         "kaniko",
 							Image:        "gcr.io/kaniko-project/executor:latest",
-							VolumeMounts: append([]v1.VolumeMount{dockerFileVolumeMount}, MakeSecretVolumeMounts(buildConfig.Secrets)...),
+							VolumeMounts: append([]v1.VolumeMount{dockerFileVolumeMount}, MakeSecretVolumeMounts([]v1.LocalObjectReference{mod.Spec.ImagePullSecret})...),
```
qbarrand, line 99: This breaks build secrets.

internal/build/job/maker.go
```diff
@@ -113,14 +113,23 @@ func (m *maker) MakeJob(mod ootov1alpha1.Module, buildConfig *ootov1alpha1.Build
...
 	for _, secretRef := range secretRefs {
 		vol := v1.Volume{
 			Name: volumeNameFromSecretRef(secretRef),
 			VolumeSource: v1.VolumeSource{
-				Secret: &v1.SecretVolumeSource{SecretName: secretRef.Name},
+				Secret: &v1.SecretVolumeSource{
+					SecretName: secretRef.Name,
+					Items: []v1.KeyToPath{
+						{
+							Key:  ".dockerconfigjson",
```
qbarrand, line 128: This function was designed to build a list of volumes referencing build secrets. This breaks an existing feature.

internal/build/job/maker.go
```diff
@@ -130,14 +139,15 @@ func MakeSecretVolumes(secretRefs []v1.LocalObjectReference) []v1.Volume {
...
+//FIXME:ybettan: if no one else uses it we should update it to a single object instead of a list
 func MakeSecretVolumeMounts(secretRefs []v1.LocalObjectReference) []v1.VolumeMount {
 	secretVolumeMounts := make([]v1.VolumeMount, 0, len(secretRefs))
 
 	for _, secretRef := range secretRefs {
 		volMount := v1.VolumeMount{
 			Name:      volumeNameFromSecretRef(secretRef),
 			ReadOnly:  true,
-			MountPath: "/run/secrets/" + secretRef.Name,
+			MountPath: "/kaniko/.docker",
```
qbarrand, line 150: Just like above, this function was designed for build secrets, not pull secrets. This breaks the [in-cluster build e2e workflow](https://github.com/qbarrand/oot-operator/blob/main/ci/module-ooto-ci-build.template.yaml#L32) which leverages build secrets.

internal/registry/registry.go
```diff
@@ -113,7 +123,48 @@ func (r *registry) ExtractToolkitRelease(layer v1.Layer) (*DriverToolkitEntry, e
...
+	secret := corev1.Secret{}
+	secretNamespacedName := types.NamespacedName{
+		Name:      ps.Name,
+		Namespace: psNamespace,
+	}
+	if err := r.client.Get(ctx, secretNamespacedName, &secret); err != nil {
+		return nil, fmt.Errorf("cannot find secret %v: %w", secretNamespacedName, err)
+	}
+
+	authData, ok := secret.Data[".dockerconfigjson"]
```
qbarrand, line 137: I think `crane` features helpers that might do all that work for us. Please have a look at https://github.com/qbarrand/oot-operator/pull/91/files#diff-896098bd06b79a03757341c930e0c577c9457ee3acd25751642a0e95782e96bfR75.

yevgeny-shnaidman:
Did not we agree that @qbarrand is going to push a new CRD with secrets both in under Build and globally?

api/v1alpha1/module_types.go
ybettan: Fixed.

ybettan:
> Did not we agree that @qbarrand is going to push a new CRD with secrets both in under Build and globally?

We did. At first I was planning to just finalize this PR and hold it until new CRD is merged. But, as it may take few more days, I can update the secret part of the CRD in this PR and then @qbarrand can just redesign all other parts of the CRD.

WDYT @yevgeny-shnaidman ?

internal/registry/registry.go
ybettan: Fixed.

internal/build/job/maker.go
ybettan: Fixed.

internal/build/job/maker.go
ybettan: Fixed.

config/crd/bases/ooto.sigs.k8s.io_modules.yaml
```diff
@@ -3954,6 +3954,19 @@ spec:
                 required:
                 - name
                 type: object
+              imagePullSecrets:
```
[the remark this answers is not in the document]
ybettan: As agreed we will have 2 kind of secrets (this PR is also making the adjustments in the API)
1. `spec.ImagePullSecret` will contains a single secret that has a single `.dockerconfigjson` file in it with the auth to all the registries required
2. `build.secrets` which will be a list of additional secrets for other uses such as cloning private github repos etc.

ybettan:
UT can still be improved but besides that PTAL.
@yevgeny-shnaidman @qbarrand @erusso7

ybettan:
@yevgeny-shnaidman @qbarrand @erusso7 PTAL.

api/v1alpha1/module_types.go
```diff
@@ -28,83 +28,82 @@ type BuildArg struct {
...
 
 	// +optional
-
 	// Push contains settings determining how to push a built DriverContainer image.
 	Push PushOptions `json:"push"`
 
-	// Secrets is an optional list of secrets to be made available to the build system.
 	// +optional
+	// Secrets is an optional list of secrets to be made available to the build system.
+	// Those secrets should be used for private resources such as a private Github repo.
```
qbarrand, line 70:
```suggestion
	// Those secrets should be used for private resources such as a GitHub access token.
```

internal/auth/auth.go
```diff
@@ -0,0 +1,47 @@
+package auth
```
qbarrand, line 1: I think `auth` is a bit too generic. WDYT about `registryauth`?

internal/auth/auth.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (a *auth) GetKeyChainFromSecret(ctx context.Context, secretName, secretNamespace string) (authn.Keychain, error) {
+
+	secret := v1.Secret{}
+	secretNamespacedName := types.NamespacedName{
+		Name:      secretName,
+		Namespace: secretNamespace,
+	}
+	if err := a.client.Get(ctx, secretNamespacedName, &secret); err != nil {
+		return nil, fmt.Errorf("cannot find secret %s/%s: %w", secretNamespace, secretName, err)
```
qbarrand, line 38:
```suggestion
		return nil, fmt.Errorf("cannot find secret %s: %w", secretNamespacedName, err)
```

internal/build/job/maker.go
```diff
@@ -113,14 +122,48 @@ func (m *maker) MakeJob(mod ootov1alpha1.Module, buildConfig *ootov1alpha1.Build
...
+					{
+						Key:  v1.DockerConfigJsonKey,
+						Path: "config.json",
+					},
+				},
+			},
+		},
+	}
+
+	return vol
```
qbarrand, line 142:
```suggestion
func (m *maker) makeImagePullSecretVolume(secretRef v1.LocalObjectReference) v1.Volume {
	return v1.Volume{
		Name: volumeNameFromSecretRef(secretRef),
		VolumeSource: v1.VolumeSource{
			Secret: &v1.SecretVolumeSource{
				SecretName: secretRef.Name,
				Items: []v1.KeyToPath{
					{
						Key:  v1.DockerConfigJsonKey,
						Path: "config.json",
					},
				},
			},
		},
	}
```

internal/build/job/maker.go
```diff
@@ -113,14 +122,48 @@ func (m *maker) MakeJob(mod ootov1alpha1.Module, buildConfig *ootov1alpha1.Build
...
+
+func (m *maker) makeImagePullSecretVolumeMount(secretRef v1.LocalObjectReference) v1.VolumeMount {
+
+	volMount := v1.VolumeMount{
+		Name:      volumeNameFromSecretRef(secretRef),
+		ReadOnly:  true,
+		MountPath: "/kaniko/.docker",
+	}
+
+	return volMount
```
qbarrand, line 153:
```suggestion
	return v1.VolumeMount{
		Name:      volumeNameFromSecretRef(secretRef),
		ReadOnly:  true,
		MountPath: "/kaniko/.docker",
	}
```

internal/build/job/maker.go
```diff
@@ -113,14 +122,48 @@ func (m *maker) MakeJob(mod ootov1alpha1.Module, buildConfig *ootov1alpha1.Build
...
+}
+
+func (m *maker) makeBuildSecretVolumes(secretRefs []v1.LocalObjectReference) []v1.Volume {
```
qbarrand, line 156: I don't think this needs to be a method.

internal/build/job/maker.go
```diff
@@ -113,14 +122,48 @@ func (m *maker) MakeJob(mod ootov1alpha1.Module, buildConfig *ootov1alpha1.Build
...
+					},
+				},
+			},
+		},
+	}
+
+	return vol
+}
+
+func (m *maker) makeImagePullSecretVolumeMount(secretRef v1.LocalObjectReference) v1.VolumeMount {
```
qbarrand, line 145: I don't think this needs to be a method.

internal/build/job/maker.go
qbarrand: Also, I don't think this should be a method.

internal/build/job/maker.go
```diff
@@ -130,7 +173,8 @@ func MakeSecretVolumes(secretRefs []v1.LocalObjectReference) []v1.Volume {
 	return volumes
 }
 
-func MakeSecretVolumeMounts(secretRefs []v1.LocalObjectReference) []v1.VolumeMount {
+func (m *maker) makeBuildSecretVolumeMounts(secretRefs []v1.LocalObjectReference) []v1.VolumeMount {
```
qbarrand, line 176: I don't think this needs to be a method.

internal/auth/auth.go
qbarrand: Or we could move it to the `registry` package. `registry.Auth` looks good to me.

internal/registry/registry.go
```diff
@@ -129,7 +135,7 @@ func (r *registry) getPullOptions(ctx context.Context, image string, po *ootov1a
 		crane.WithContext(ctx),
 	}
 
-	if po != nil {
+	if *po != (ootov1alpha1.PullOptions{}) {
```
yevgeny-shnaidman, line 138: why not just compare to nil?

ybettan: Because `PullOption` isn't a pointer (in its definition context), which means that if it wasn't set by the user, its default value will be an empty struct, and when passed by reference to `getPullOptions` it will get the value of `&ootov1alpha1.PullOptions{}` and not `nil`.

I do see though that we are sometimes calling this function with the explicit `nil` arg so I will change the condition to `if po != nil && *po != (ootov1alpha1.PullOptions{})`.

api/v1alpha1/module_types.go
```diff
@@ -121,10 +120,14 @@ type ModuleSpec struct {
...
 
+	// +optional
 	// ServiceAccountName is the name of the ServiceAccount to use to run this pod.
 	// More info: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
-	// +optional
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+
+	// +optional
+	// ImagePullSecret is the name of the single Secret to use for all container image registries that requires auth.
+	ImagePullSecret v1.LocalObjectReference `json:"imagePullSecret,omitempty"`
```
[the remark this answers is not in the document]
ybettan: Good idea, will do it.

internal/build/job/maker.go
```diff
@@ -78,6 +78,15 @@ func (m *maker) MakeJob(mod ootov1alpha1.Module, buildConfig *ootov1alpha1.Build
 		MountPath: "/workspace",
 	}
 
+	volumes := []v1.Volume{dockerFileVolume}
+	volumeMounts := []v1.VolumeMount{dockerFileVolumeMount}
+	if mod.Spec.ImagePullSecret != (v1.LocalObjectReference{}) {
```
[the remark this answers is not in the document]
ybettan: Good point, will do.

internal/auth/auth.go
ybettan: I liked the fact that `auth` is a bit generic so it can be reused and extended also in other contexts such as interacting with certs configMaps in the future.

For this reason I also would like to to make it a nested package of the `registry` pkg.

internal/auth/auth.go
ybettan: I though it is a bit more explicit with %s/%s but I don't mind so I will change it if you prefer it :)

internal/build/job/maker.go
ybettan: Agree. Fixing.

internal/build/job/maker.go
ybettan: Agree. Fixing.

internal/build/job/maker.go
ybettan: Agree. Fixing.

internal/build/job/maker.go
ybettan: Agree. Fixing.

internal/build/job/maker.go
ybettan: Agree. Fixing.

ybettan:
@qbarrand @yevgeny-shnaidman PTAL.

internal/registry/registry.go
yevgeny-shnaidman: What i mean is that even default values of PullOptions are still valid values. So we can just check the pointer not nil, and proceed to update according to the default values. If in the future default values will have to be used for some setting, comparing to ootov1alpha1.PullOptions{}) will cause buggy flow.

ybettan: I see what you mean. Fixed it as well. Thanks.

internal/auth/auth.go
qbarrand: > I liked the fact that `auth` is a bit generic

We're only talking about authentication against a registry, right? I don't think we would use push secrets for any other purpose.

> it can be reused and extended also in other contexts such as interacting with certs configMaps in the future

You mean CA certificates? Those are not about authenticating but securing the connection to the registry.

> For this reason I also would not like to to make it a nested package of the `registry` pkg.

I think it would be very appropriate for this code to be there.

yevgeny-shnaidman (approved):

internal/auth/auth.go
qbarrand: I think it makes sense to rely on the `NamespacedName`'s `String` here.

ybettan: Done.

internal/auth/auth.go
ybettan: As agreed, we will keep it as an `auth` internal package, and the interface will be renamed to `RegistryAuth`. In the future, this package can be extended with additional interfaces such as `GithubAuth` etc.

PR #76, opened 2022-06-16
Support in-cluster build overrides

yevgeny-shnaidman:
This PR adds support for in-cluster build overrides on the
kernel mapping basis. In case there is a general build config defined
in the Spec, and the build config defined in the kernel mapping,
then some of the fields of the KM build config will override
the fields in the general spec (Dockerfile), and other fileds will be concatinated
( BuildArgs, Secrets)

MGMT-10539


controllers/build/modulehelper.go
```diff
@@ -17,7 +20,7 @@ func NewModuleHelper() ModuleHelper {
 	return moduleHelper{}
 }
 
-func (moduleHelper) ApplyBuildArgOverrides(args []v1alpha1.BuildArg, overrides ...v1alpha1.BuildArg) []v1alpha1.BuildArg {
+func (m moduleHelper) ApplyBuildArgOverrides(args []v1alpha1.BuildArg, overrides ...v1alpha1.BuildArg) []v1alpha1.BuildArg {
```
ybettan, line 23: This method doesn't use the `moduleHelper` obj. Should we make it a function instead of a method?

controllers/build/modulehelper.go
```diff
@@ -41,3 +44,26 @@ func (moduleHelper) ApplyBuildArgOverrides(args []v1alpha1.BuildArg, overrides .
 
 	return args
 }
+
+func (m moduleHelper) GetReleventBuild(mod ootov1alpha1.Module, km ootov1alpha1.KernelMapping) *ootov1alpha1.Build {
```
ybettan, line 48: This method doesn't use the `moduleHelper` obj. Should we make it a function instead of a method?

ybettan:
If I understood correctly, each time we need to handle a build we check the higher layer and to take global configs and then override them with the relevant configs of the kernelMapper.

Thant means that in each place in the code (and also future references) that need the build object will need to repeat that same logic.

I would like to suggest another option.
How about having a logic at the beginning of the reconcile loop that import all `module.build` data to `module.kernelMapping[].build` (excluding the one overwritten by the `kernelMapping`)

That way, in everyplace in the code, we only need to look at `module.kernelMapping[].build` and assume it is already complete.

The same design can be applied to other fields that are inheriting from a higher layer such as:
* `module.kernelMapping.containerImage` inheriting from `module.container.image`
* future global config

controllers/build/modulehelper.go
yevgeny-shnaidman: I always prefer using methods instead of module functions. Makes code clear and, in addition, makes it easier to write unit tests

controllers/build/modulehelper.go
yevgeny-shnaidman: I always prefer using methods instead of module functions. Makes code clear and, in addition, makes it easier to write unit tests

yevgeny-shnaidman:
> If I understood correctly, each time we need to handle a build we check the higher layer and to take global configs and then override them with the relevant configs of the kernelMapper.
> 
> Thant means that in each place in the code (and also future references) that need the build object will need to repeat that same logic.
> 
> I would like to suggest another option. How about having a logic at the beginning of the reconcile loop that import all `module.build` data to `module.kernelMapping[].build` (excluding the one overwritten by the `kernelMapping`)
> 
> That way, in everyplace in the code, we only need to look at `module.kernelMapping[].build` and assume it is already complete.
> 
> The same design can be applied to other fields that are inheriting from a higher layer such as:
> 
> * `module.kernelMapping.containerImage` inheriting from `module.container.image`
> * future global config

We will need the unified build object only in case we do need to build the image. And right now we are getting the unified build config at the beginning of the build flow, and then just pass it down to the functions

ybettan (approved):

PR #80, opened 2022-06-19
Enabling metrics for KMMO

yevgeny-shnaidman:
This PR prepares the infrastructure for metrics for KMMO
and add the first initial metric: number of KMMO CRs in
the cluster

MGMT-10543


controllers/module_reconciler.go
```diff
@@ -45,10 +46,11 @@ import (
...
 
-	bm build.Manager
-	dc DaemonSetCreator
-	km module.KernelMapper
-	su module.ConditionsUpdater
+	bm     build.Manager
+	dc     DaemonSetCreator
+	km     module.KernelMapper
+	su     module.ConditionsUpdater
+	metric metrics.Metrics
```
ybettan, line 53:
```suggestion
	metrics metrics.Metrics
```

controllers/module_reconciler.go
```diff
@@ -82,10 +85,8 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	logger := log.FromContext(ctx)
 
-	mod := ootov1alpha1.Module{}
-
-	if err := r.Client.Get(ctx, req.NamespacedName, &mod); err != nil {
-		logger.Error(err, "Could not get module")
+	mod, err := r.getRequestedModule(ctx, req.NamespacedName)
+	if err != nil {
 		return res, err
```
ybettan, line 90: We should probably wrap the error here.

controllers/module_reconciler.go
```diff
@@ -262,6 +263,27 @@ func (r *ModuleReconciler) handleDevicePlugin(ctx context.Context, mod *ootov1al
...
 
+func (r *ModuleReconciler) getRequestedModule(ctx context.Context, namespacedName types.NamespacedName) (*ootov1alpha1.Module, error) {
+	logger := log.FromContext(ctx)
+
+	mods := ootov1alpha1.ModuleList{}
+
+	err := r.Client.List(context.Background(), &mods)
+	if err != nil {
+		logger.Error(err, "could not list modules")
+		return nil, err
```
ybettan, line 274: Should we wrap the error here?

controllers/module_reconciler.go
```diff
@@ -262,6 +263,27 @@ func (r *ModuleReconciler) handleDevicePlugin(ctx context.Context, mod *ootov1al
...
+	}
+
+	r.metric.SetExistingKMMOsCRs(len(mods.Items))
```
ybettan, line 277:
```suggestion
	r.metric.SetExistingKModuleCRs(len(mods.Items))
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,44 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	runtimemetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+// When adding metric names, see https://prometheus.io/docs/practices/naming/#metric-names
+const (
+	existingKMMOCRsQuery = "kmmo_number_crs_total"
```
ybettan, line 10:
```suggestion
	existingModuleCRsQuery = "module_number_crs_total"
```

I think you are only listing the `Module` objects.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,44 @@
...
+	Register()
+	SetExistingKMMOsCRs(value int)
+}
+
+func New() Metrics {
+
+	kmmoResourcesNum := prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: existingKMMOCRsQuery,
+			Help: "Number of existing KMMO CRs",
```
ybettan, line 26:
```suggestion
			Help: "Number of existing Module CRs",
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,44 @@
...
+	return &metrics{kmmoResourcesNum: kmmoResourcesNum}
+}
+
+func (m *metrics) Register() {
+	runtimemetrics.Registry.MustRegister(
+		m.kmmoResourcesNum,
+	)
+}
+
+type metrics struct {
```
ybettan, line 38: Can we move that declaration just beneath `Metrics` interface just for convenience?

controllers/module_reconciler.go
yevgeny-shnaidman: fixed

controllers/module_reconciler.go
yevgeny-shnaidman: not sure, it is wrapped and pretty much explained in the called function, we are not going to get any new info from wrapping it

controllers/module_reconciler.go
yevgeny-shnaidman: fixed

controllers/module_reconciler.go
yevgeny-shnaidman: isn't KMMO better? That's what we call the operator, so it should handle KMMO CRs. Not sure what does it mean KModule

pkg/metrics/metrics.go
yevgeny-shnaidman: Same comment here. Also it it better to keep a prefix of kmmo for all the metrix, will make it much easier to find and group them

pkg/metrics/metrics.go
yevgeny-shnaidman: Same here

controllers/module_reconciler.go
ybettan: IMO this is not just for "adding info" but also to make it easier to track the error in the code base and understand the exact failure point instead of guessing.

controllers/module_reconciler.go
ybettan: In `module_reconciler.go` you do `err := r.Client.List(context.Background(), &mods)` which mean you only list `Modules` resources. We may have more CustomResources in the future (preflight for example).

By using `KMMOsCRs` I expect it to count all CRs (not only modules) but this is not what the code you added is doing so I think it is missleading don't you think so?!

pkg/metrics/metrics.go
ybettan: With that I agree. How about `kmmo_module_number_crs_total`?

config/prometheus/monitor.yaml
```diff
@@ -18,3 +18,32 @@ spec:
...
+metadata:
+  name: prometheus-k8s
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+  - kind: ServiceAccount
+    name: prometheus-k8s
+    namespace: openshift-monitoring
```
qbarrand, line 49: This is OpenShift-specific - should we wait until we have a downstream repo before we add this file?

config/manager/manager.yaml
```diff
@@ -3,6 +3,7 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller-manager
+    openshift.io/cluster-monitoring: "true"
```
qbarrand, line 6: OpenShift-specific label.

controllers/module_reconciler.go
```diff
@@ -262,6 +263,24 @@ func (r *ModuleReconciler) handleDevicePlugin(ctx context.Context, mod *ootov1al
...
 
+func (r *ModuleReconciler) getRequestedModule(ctx context.Context, namespacedName types.NamespacedName) (*ootov1alpha1.Module, error) {
+	mods := ootov1alpha1.ModuleList{}
+
+	err := r.Client.List(context.Background(), &mods)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list modules in the cluster: %w", err)
+	}
+
+	r.metrics.SetExistingKMMOsCRs(len(mods.Items))
```
qbarrand, line 274: Wondering if it is the right place for this call. `getRequestedModule` sounds like a utility function; although it's cheap, updating the metric is a side-effect.

controllers/module_reconciler.go
```diff
@@ -262,6 +263,24 @@ func (r *ModuleReconciler) handleDevicePlugin(ctx context.Context, mod *ootov1al
...
+	for i, item := range mods.Items {
+		if item.Name == namespacedName.Name && item.Namespace == namespacedName.Namespace {
+			return &mods.Items[i], nil
+		}
+	}
+
+	return nil, fmt.Errorf("failed to find the module %s/%s", namespacedName.Namespace, namespacedName.Name)
```
qbarrand, line 281:
```suggestion
	return nil, fmt.Errorf("failed to find module %s", namespacedName)
```

pkg/metrics/metrics.go
qbarrand: Aren't you removing the `kmmo` prefix here?

config/prometheus/monitor.yaml
yevgeny-shnaidman: i added it right now, so that we won't forget it ( also to allow checking on OCP). in ant case, the prometheus is commented in the default kustomezation, so it is not deployed on the K8s

controllers/module_reconciler.go
yevgeny-shnaidman: i prefered to put it there , so as not to return the number of KMMO in the cluster, because it also sound as a side-effect to actually finding the right module

controllers/module_reconciler.go
yevgeny-shnaidman: fixed

pkg/metrics/metrics.go
yevgeny-shnaidman: module is a the second "m" in kmmo

controllers/module_reconciler.go
yevgeny-shnaidman: preflight won't be a Module CR , it will be a different one, so that won't be a problem

controllers/module_reconciler.go
yevgeny-shnaidman: fixed

config/manager/manager.yaml
yevgeny-shnaidman: fixed

controllers/module_reconciler.go
yevgeny-shnaidman: fixed

controllers/module_reconciler.go
```diff
@@ -82,13 +85,13 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	logger := log.FromContext(ctx)
 
-	mod := ootov1alpha1.Module{}
-
-	if err := r.Client.Get(ctx, req.NamespacedName, &mod); err != nil {
-		logger.Error(err, "Could not get module")
-		return res, err
+	mod, err := r.getRequestedModule(ctx, req.NamespacedName)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: better to have it in a separate function: makes code easier to read, and makes changing the implementation easier, in case we want to add some functionality (check) to the function

qbarrand (approved):

ybettan (approved):

PR #82, opened 2022-06-23
Refactor operator code, split it to internal packages

yevgeny-shnaidman:
This PR refactors the code base so that the repo
will have 2 main directories:
1) controllers - will contain almost exclusively controller code
2) internal - will contain all the packages used by the controllers' code

MGMT-10621


Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN go mod download
 COPY main.go main.go
 COPY api api
 COPY controllers controllers
+COPY internal internal
 COPY pkg pkg
```
ybettan, line 19: We can probably remove `pkg` from the container image.

Also we should probably move all the files in `pkg` to another location.

ybettan:
The `controller/daemonset` pkg isn't a controller but rather an internal package. Shouldn't it be moved to `internal/` as well?

controllers/daemonset.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"github.com/go-openapi/swag"
 	ootov1alpha1 "github.com/qbarrand/oot-operator/api/v1alpha1"
-	"github.com/qbarrand/oot-operator/controllers/constants"
```
ybettan, line 10: The controller/daemonset pkg isn't a controller but rather an internal package. Shouldn't it be moved to internal/ as well?

internal/build/helper.go
```diff
@@ -7,20 +7,20 @@ import (
...
 	ApplyBuildArgOverrides(args []v1alpha1.BuildArg, overrides ...v1alpha1.BuildArg) []v1alpha1.BuildArg
 	GetReleventBuild(mod ootov1alpha1.Module, km ootov1alpha1.KernelMapping) *ootov1alpha1.Build
 }
 
-type moduleHelper struct{}
+type helper struct{}
 
-func NewModuleHelper() ModuleHelper {
-	return moduleHelper{}
+func NewHelper() Helper {
```
ybettan, line 19: I think every `New*` function in the code should return a struct and not an interface.
The "thumb rule" in golang says "accept interfaces, return structs"

Here are a few reasons for that:
* By returning an interface, we restrict the user of the package to access public fields in the struct
* The "user" should be able to decide if the returned object is stored as an interface or as a struct (more flexibility). In case the user only want the interface as return value, we should use `var helper Helper = NewHelper()` but not restrict it at the library level.
* What will be returned in case an object implement multiple interfaces (even if we don't have such in the code ATM)? This is not working when returning interfaces.

internal/build/helper.go
```diff
@@ -7,20 +7,20 @@ import (
...
+	return helper{}
```
ybettan, line 20:
```suggestion
	return &helper{}
```
?

internal/build/helper.go
```diff
@@ -7,20 +7,20 @@ import (
...
 }
 
-func (m moduleHelper) ApplyBuildArgOverrides(args []v1alpha1.BuildArg, overrides ...v1alpha1.BuildArg) []v1alpha1.BuildArg {
+func (m helper) ApplyBuildArgOverrides(args []v1alpha1.BuildArg, overrides ...v1alpha1.BuildArg) []v1alpha1.BuildArg {
```
ybettan, line 23:
```suggestion
func (m *helper) ApplyBuildArgOverrides(args []v1alpha1.BuildArg, overrides ...v1alpha1.BuildArg) []v1alpha1.BuildArg {
```

And same for all other methods of `helper`.

internal/build/job/maker.go
```diff
@@ -19,18 +19,18 @@ type Maker interface {
...
 
 type maker struct {
-	mh     build.ModuleHelper
+	helper build.Helper
 	scheme *runtime.Scheme
 }
 
-func NewMaker(mh build.ModuleHelper, scheme *runtime.Scheme) Maker {
-	return &maker{mh: mh, scheme: scheme}
+func NewMaker(helper build.Helper, scheme *runtime.Scheme) Maker {
```
ybettan, line 26:
```suggestion
func NewMaker(helper build.Helper, scheme *runtime.Scheme) *maker {
```
See https://github.com/qbarrand/oot-operator/pull/82/files#r906702457

internal/filter/filter.go
```diff
@@ -0,0 +1,123 @@
...
+	NodeKernelReconcilerPredicate(labelName string) predicate.Predicate
+	FindModulesForNode(node client.Object) []reconcile.Request
+}
+
+type filter struct {
+	client client.Client
+	logger logr.Logger
+}
+
+func New(client client.Client, logger logr.Logger) Filter {
```
ybettan, line 31:
```suggestion
func New(client client.Client, logger logr.Logger) *filter {
```

internal/filter/filter.go
```diff
@@ -0,0 +1,123 @@
...
+
+/*
+func Namespace(namespace string) predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.GetNamespace() == namespace
+	})
+}
+*/
+
+var skipDeletions predicate.Predicate = predicate.Funcs{
```
ybettan, line 121: Do we have to user a global variable here? Can it be part of the `filter` object instead?
If not, can we move it at the top of the file as
```
var (
    ...
)
```
?

internal/filter/filter.go
```diff
@@ -0,0 +1,123 @@
...
+	})
+}
+
+/*
+func Namespace(namespace string) predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.GetNamespace() == namespace
+	})
+}
+*/
```
ybettan, line 119: Can we remove it if we don't use it?

internal/filter/filter_test.go
```diff
@@ -0,0 +1,293 @@
...
+				Namespace(targetNamespace).Delete(ev),
+			).To(
+				Equal(expected),
+			)
+		},
+		Entry("matching namespace", "ns", "ns", true),
+		Entry("different namespaces", "ns", "other-ns", false),
+	)
+})
+*/
```
ybettan, line 67: Remove?

internal/module/conditions.go
```diff
@@ -10,9 +10,9 @@ import (
 )
 
 const (
```
ybettan, line 12: Can we change the package name from `module` to `conditions`? I think `module` is a bit confusing here.

internal/module/conditions_test.go
```diff
@@ -1,12 +1,11 @@
-package module_test
+package module
```
ybettan, line 1: Can we change the package name from module to conditions? I think module is a bit confusing here.

main.go
```diff
@@ -119,11 +121,13 @@ func main() {
 
 	setupLogger.V(1).Info("Determining kernel labeling method", KernelLabelingMethodEnvVar, kernelLabelingMethod)
 
+	filterAPI := filter.New(client, mgr.GetLogger())
```
ybettan, line 124:
```suggestion
	var filterAPI filter.Filter = filter.New(client, mgr.GetLogger())
```

main.go
```diff
@@ -119,11 +121,13 @@ func main() {
...
+
 	switch kernelLabelingMethod {
 	case OOTOKernelLabelingMethod:
 		kernelLabel = "oot.node.kubernetes.io/kernel-version.full"
 
-		nkr := controllers.NewNodeKernelReconciler(client, kernelLabel)
+		nkr := controllers.NewNodeKernelReconciler(client, kernelLabel, filterAPI)
```
ybettan, line 130:
```suggestion
		nodeKernelReconciler := controllers.NewNodeKernelReconciler(client, kernelLabel, filterAPI)
```
?

ybettan:
Although I have a few comments I like the content of this PR very much!

controllers/daemonset.go
yevgeny-shnaidman: i thought about that , but could not find where to move it:  internal/daemonset?

Dockerfile
yevgeny-shnaidman: i was going to leave to @erusso7 , since he is going to introduce client pkg, which will naturally go into internal and will allow smooth removal if test

internal/build/helper.go
yevgeny-shnaidman: I agree with the points that you mentioned here, and will change the code, but there are 3 counter-arguments:
1) by returning interface you are restricting the usage of your package. User can only use the package through interface that you defined. That, in turn, makes it safer for you to change the implementation code without worrying that you can break user code
2) by returning interface m you are actually checking that your stuct has implemented all the functions defined by the interface during compilation time. Otherwise it will be checked only when user code is compiled, and only in case he uses it as an interface
3) it is not relevant for our code, but in case we return pointer instead of interface, the inner structure must be made public, otherwise the variable cannot be defined explicitly

internal/filter/filter.go
yevgeny-shnaidman: fixed, moved vars to the top of the file

internal/filter/filter.go
yevgeny-shnaidman: fixed

internal/filter/filter_test.go
yevgeny-shnaidman: fixed

main.go
yevgeny-shnaidman: why do you want to define it explicitly? We are just passing all the APIs to the manager/other functions, without actually using them

main.go
yevgeny-shnaidman: fixed

internal/filter/filter.go
yevgeny-shnaidman: fixed

internal/build/job/maker.go
yevgeny-shnaidman: fixed

internal/build/helper.go
yevgeny-shnaidman: fixed

internal/build/helper.go
yevgeny-shnaidman: fixed

controllers/daemonset.go
ybettan: > internal/daemonset

Yes, why not (how does it differ from other packages moved to `internal/`)?

internal/build/helper.go
ybettan: Thanks. Just for more clarification:

> by returning interface you are restricting the usage of your package. User can only use the package through interface that you defined. That, in turn, makes it safer for you to change the implementation code without worrying that you can break user code

Isn't that what private methods/fields are for? By setting a public field, this is treated as API and shouldn't be changed (same as we "don't" change interface methods).

> by returning interface m you are actually checking that your stuct has implemented all the functions defined by the interface during compilation time. Otherwise it will be checked only when user code is compiled, and only in case he uses it as an interface

+1 on that. Although I think that UT should cover this one but it is indeed better to have the compiler check it.

> it is not relevant for our code, but in case we return pointer instead of interface, the inner structure must be made public, otherwise the variable cannot be defined explicitly

Actually this one is a very good point!

main.go
ybettan: This one is a nit. I am OK with it either way.

controllers/daemonset.go
yevgeny-shnaidman: fixed

internal/metrics/metrics.go
```diff
@@ -22,7 +22,7 @@ type metrics struct {
 	kmmoResourcesNum prometheus.Gauge
 }
 
-func New() Metrics {
+func New() *metrics {
```
ybettan, line 25: Sorry. Another small one. Do we want to have a more meaningful name? Let say `NewMetrics`?

yevgeny-shnaidman: I am not sure about it.  I think that package name allows to uniquely identify the function. The problem arises when we have more then one interface/struct in the same package

internal/module/conditions.go
```diff
@@ -10,9 +10,9 @@ import (
 )
```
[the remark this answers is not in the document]
yevgeny-shnaidman: actually i don't remember exactly what we have agreed upon. I suggest we leave it here for now, and it will be removed once the condition/status implementation of the CR will be delivered

controllers/module_reconciler.go
```diff
@@ -38,35 +38,37 @@ import (
...
-	dc      DaemonSetCreator
-	km      module.KernelMapper
-	su      module.ConditionsUpdater
-	metrics metrics.Metrics
+	buildAPI      build.Manager
+	daemonAPI     DaemonSetCreator
+	kernelAPI     module.KernelMapper
+	conditionsAPI module.ConditionsUpdater
+	metricsAPI    metrics.Metrics
+	filterAPI     filter.Filter
```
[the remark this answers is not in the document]
yevgeny-shnaidman: i think it is an easy way to tell that those are actually interfaces...

internal/filter/filter.go
```diff
@@ -0,0 +1,123 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+//go:generate mockgen -source=filter.go -package=filter -destination=mock_filter_api.go
+
+type Filter interface {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/build/helper.go
```diff
@@ -45,7 +45,7 @@ func (m moduleHelper) ApplyBuildArgOverrides(args []v1alpha1.BuildArg, overrides
 	return args
 }
 
-func (m moduleHelper) GetReleventBuild(mod ootov1alpha1.Module, km ootov1alpha1.KernelMapping) *ootov1alpha1.Build {
+func (m *helper) GetReleventBuild(mod ootov1alpha1.Module, km ootov1alpha1.KernelMapping) *ootov1alpha1.Build {
```
qbarrand, line 48:
```suggestion
func (m *helper) GetRelevantBuild(mod ootov1alpha1.Module, km ootov1alpha1.KernelMapping) *ootov1alpha1.Build {
```

internal/daemonset/daemonset_test.go
```diff
@@ -26,7 +25,7 @@ const (
 )
 
 var _ = Describe("SetDriverContainerAsDesired", func() {
-	dg := controllers.NewDaemonSetCreator(nil, kernelLabel, scheme)
```
qbarrand, line 29: If we now have a `daemonset` package, maybe this could be renamed as `daemonset.NewCreator`?

qbarrand (commented):
Last nits, otherwise LGTM 🙂

internal/build/helper.go
yevgeny-shnaidman: fixed

internal/build/job/maker.go
```diff
@@ -19,18 +19,18 @@ type Maker interface {
...
 
 type maker struct {
-	mh     build.ModuleHelper
+	helper build.Helper
 	scheme *runtime.Scheme
 }
 
-func NewMaker(mh build.ModuleHelper, scheme *runtime.Scheme) Maker {
-	return &maker{mh: mh, scheme: scheme}
+func NewMaker(helper build.Helper, scheme *runtime.Scheme) *maker {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: see this discussion:
https://github.com/qbarrand/oot-operator/pull/82#discussion_r906702457

internal/daemonset/daemonset_test.go
yevgeny-shnaidman: fixed

ybettan (approved):

qbarrand (approved):

PR #85, opened 2022-06-27
Adding documentation for pre-building driver-containers.

ybettan:
KMMO is mounting `/lib/modules/$(KVER)` from the node to the driver-container
in order to use potential in-tree dependencies, therefore, some adjustment
are needed for the driver-container so KMMO don't override the *.ko files of
the user.

MGMT-10194


docs/mkdocs/driver_containers.md
```diff
@@ -0,0 +1,12 @@
+# How to build driver-containers to be used by KMMO
+
+### Pre-build driver-containers
+
+When a `Module` CR is applied to the cluster, KMMO will create a DaemonSet to run the driver-container specified in it.
+
+In order for driver-container that depends on in-tree kernel-modules to work, KMMO is mounting `/lib/modules/$(KVER)` from the node to the container, which means that it will override all files located in `/lib/modules/$(KVER)` inside the driver-container.
```
qbarrand, line 7:
```suggestion
In order for driver-containers that depend on in-tree kernel-modules to work, KMMO is mounting `/lib/modules/${KVER}` from the node to the container, which means that it will override all files located in `/lib/modules/${KVER}` inside the driver-container.
```

docs/mkdocs/driver_containers.md
```diff
@@ -0,0 +1,12 @@
...
+
+Therefore, when building a dirver-container that is going to be used by KMMO, there are few things that needs to be done in the driver-container:
```
qbarrand, line 9:
```suggestion
Therefore, when building a driver-container that is going to be used by KMMO, there are few things that needs to be done in the driver-container:
```

docs/mkdocs/driver_containers.md
```diff
@@ -0,0 +1,12 @@
...
+* Put the  `*.ko` files in `/opt/lib/modules/$(KVER)` instead of `/lib/modules/$(KVER)`
+* Link `/lib/modules/$(KVER)` inside `/opt/lib/modules/$(KVER)/system` in case the driver-container depend on in-tree kernel-modules
+* Run `depmod -b /opt` in order to generate the dependency file correctly
```
qbarrand, line 12:
```suggestion
* Put the `*.ko` files in `/opt/lib/modules/${KVER}` instead of `/lib/modules/${KVER}`
* Link `/lib/modules/${KVER}` inside `/opt/lib/modules/$(KVER)/system` in case the driver-container depend on in-tree kernel-modules
* Run `depmod -b /opt` in order to generate the dependency file correctly
```

PR #90, opened 2022-06-30
Adding the possibility to run a single-package in unit-tests.

ybettan:
In order to run only the `filter` package tests for example, use:
```
make unit-test TEST=github.com/qbarrand/oot-operator/internal/filter
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,9 +107,11 @@ fmt: ## Run go fmt against code.
 vet: ## Run go vet against code.
 	go vet ./...
 
+TEST ?= ./...
+
 .PHONY: test
 unit-test: manifests generate lint vet envtest ## Run tests.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -51,6 +51,8 @@ IMG ?= ghcr.io/qbarrand/oot-operator:main
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.23
 
+TEST ?= ./...
```
qbarrand, line 54: Could we have that variable defined just before the `unit-test` target?

ybettan: Done.

qbarrand (approved):

PR #92, opened 2022-07-01
Use Background context instead of TODO

erusso7:
Following the good practices of Golang,
we should use context.Background() function in test files
instead of context.TODO()


qbarrand (approved):
Thank you!

PR #93, opened 2022-07-01
Replacing registry internal package implementation

yevgeny-shnaidman:
This PR partky import registry package from SRO V1. It will
now support the following API
1) ImageExists - used by build
2) ExtractToolkitRelease - will be used by preflight
3) GetLayersDigests - will be used by preflight
4) GetLayerByDigest - will be used by preflight

MGMT-10829


internal/build/job/manager_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/qbarrand/oot-operator/internal/build"
 	"github.com/qbarrand/oot-operator/internal/client"
 	"github.com/qbarrand/oot-operator/internal/constants"
-	"github.com/qbarrand/oot-operator/internal/registry"
+	registrypkg "github.com/qbarrand/oot-operator/internal/registry"
```
ybettan, line 14: Why not just `registry`?

internal/registry/registry.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func (r *registry) ImageExists(ctx context.Context, image string, po ootov1alpha1.PullOptions) (bool, error) {
+	_, _, err := r.getImageManifest(ctx, image, &po)
+	if err != nil {
+		te := &transport.Error{}
+		if errors.As(err, &te) && te.StatusCode == http.StatusNotFound {
+			return false, nil
+		}
+		return false, fmt.Errorf("could not get image: %w", err)
```
ybettan, line 61:
```suggestion
		return false, fmt.Errorf("could not get image %s: %w", image, err)
```

internal/registry/registry.go
```diff
@@ -0,0 +1,242 @@
...
+	return digests, pullConfig, nil
+}
+
+func (r *registry) GetLayerByDigest(digest string, pullConfig *repoPullConfig) (v1.Layer, error) {
+	return crane.PullLayer(pullConfig.repo+"@"+digest, pullConfig.authOptions...)
+}
+
+func (r *registry) ExtractToolkitRelease(layer v1.Layer) (*DriverToolkitEntry, error) {
+	var err error
+	var found bool
```
ybettan, line 86:
```suggestion
        var (
	    err error
	    found bool
	)
```

internal/registry/registry.go
```diff
@@ -0,0 +1,242 @@
...
+	dtk := &DriverToolkitEntry{}
+	obj, err := r.getHeaderFromLayer(layer, driverToolkitJSONFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find file %s in image layer: %w", driverToolkitJSONFile, err)
+	}
+
+	dtk.KernelFullVersion, found, err = unstructured.NestedString(obj.Object, "KERNEL_VERSION")
+	if !found || err != nil {
```
ybettan, line 94: Isn't it enough to only check `err != nil`?

internal/registry/registry.go
```diff
@@ -0,0 +1,242 @@
+package registry
```
ybettan, line 1: I think we shouldn't merge it without UTs

internal/build/job/manager_test.go
yevgeny-shnaidman: because we have a variable named registry, that is used in multiple places

internal/registry/registry.go
yevgeny-shnaidman: fixed

internal/registry/registry.go
yevgeny-shnaidman: fixed

internal/registry/registry.go
yevgeny-shnaidman: this is a hard flow to check in unit-test: you need to create repo and images. not worth it for unit-test. we also did not have it in V1

ybettan: Ack. I have talked with @qbarrand and this package is probably a good candidate for subsystem-tests when we have the occasion.

yevgeny-shnaidman: agree

internal/registry/registry.go
```diff
@@ -0,0 +1,241 @@
...
+
+	digests, err := r.getLayersDigestsFromManifestStream(manifest)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get layers digests from manifest of the image %s: %w", image, err)
+	}
+
+	return digests, pullConfig, nil
+}
+
+func (r *registry) GetLayerByDigest(digest string, pullConfig *repoPullConfig) (v1.Layer, error) {
```
qbarrand, line 80: This is a public method, but it accepts an unexported type as parameter (`*repoPullConfig`).

internal/registry/registry.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	dtk.OSVersion, found, err = unstructured.NestedString(obj.Object, "RHEL_VERSION")
+	if !found || err != nil {
+		return nil, fmt.Errorf("failed to get RHEL_VERSION from %s, found %t: %w", driverToolkitJSONFile, found, err)
+	}
+	return dtk, nil
+}
+
+func (r *registry) getImageManifest(ctx context.Context, image string, po *ootov1alpha1.PullOptions) ([]byte, *repoPullConfig, error) {
```
qbarrand, line 109: I was already confused in v1 why this method returns the pull configuration i.e. the crane options.
Could we extract that code into a separate function that only generates the options?

internal/registry/registry.go
```diff
@@ -0,0 +1,241 @@
...
+				return nil, fmt.Errorf("failed to unmarshal tar entry: %w", err)
+			}
+			return &obj, nil
+		}
+	}
+
+	return nil, fmt.Errorf("header %s not found in the layer", headerName)
+}
+
+func (r *registry) getImageDigestFromMultiImage(manifestListStream []byte) (string, error) {
```
qbarrand, line 228: This one looks unit testable, WDYT?

internal/registry/registry.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	digests := make([]string, len(manifest.Layers))
+	for i, layer := range manifest.Layers {
+		digests[i] = layer.Digest.Algorithm + ":" + layer.Digest.Hex
+	}
+	return digests, nil
+}
+
+func (r *registry) getHeaderFromLayer(layer v1.Layer, headerName string) (*unstructured.Unstructured, error) {
```
qbarrand, line 183: This one looks unit testable, WDYT?
Also, should we return a better / more concrete type as the header?

internal/registry/registry.go
```diff
@@ -0,0 +1,241 @@
...
+		// get the manifest stream for the image of the architecture
+		manifest, err = crane.Manifest(repo+"@"+archDigest, options...)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get crane manifest for the arch image: %w", err)
+		}
+	}
+	return manifest, nil
+}
+
+func (r *registry) getLayersDigestsFromManifestStream(manifestStream []byte) ([]string, error) {
```
qbarrand, line 169: This function looks unit testable.

internal/registry/registry.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	return digests, pullConfig, nil
+}
+
+func (r *registry) GetLayerByDigest(digest string, pullConfig *repoPullConfig) (v1.Layer, error) {
+	return crane.PullLayer(pullConfig.repo+"@"+digest, pullConfig.authOptions...)
+}
+
+func (r *registry) ExtractToolkitRelease(layer v1.Layer) (*DriverToolkitEntry, error) {
```
qbarrand, line 84: I would keep that only for the downstream version.

internal/registry/registry.go
yevgeny-shnaidman: fixed

internal/registry/registry.go
yevgeny-shnaidman: as agreed, this package will be fully tested in the sub-system

internal/registry/registry.go
yevgeny-shnaidman: Why? even in upstream we will define that image must contain drtvier toolkit release file in order for preflight to work. Maybe the name of the file(and function) should be changed, but the meaning stays the same

internal/registry/registry.go
yevgeny-shnaidman: fixed

qbarrand (approved):

internal/registry/registry.go
```diff
@@ -0,0 +1,250 @@
...
+)
+
+type DriverToolkitEntry struct {
+	ImageURL            string `json:"imageURL"`
+	KernelFullVersion   string `json:"kernelFullVersion"`
+	RTKernelFullVersion string `json:"RTKernelFullVersion"`
+	OSVersion           string `json:"OSVersion"`
+}
+
+// [TODO] make this struct public, once there is a code that is going to use it
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

PR #95, opened 2022-07-04
Use k8s.io/utils/pointer to create pointer values in Kubernetes objects

qbarrand:
To add pointer values to Kubernetes objects, we were either using an OpenAPI library or manually declaring variables. This PR standardizes around the `k8s.io/utils/pointer` package to generate those values.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	k8s.io/apimachinery v0.24.2
 	k8s.io/client-go v0.24.2
 	k8s.io/klog/v2 v2.70.0
+	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9
 	sigs.k8s.io/controller-runtime v0.12.2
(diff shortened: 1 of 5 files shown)


yevgeny-shnaidman:
why not continue using swag? Does te same, only has a lot more options

qbarrand:
@yevgeny-shnaidman I did not give it much thought - I discovered the Kubernetes library by accident and thought it would make sense to standardize on something coming from the Kubernetes org. Maybe we can use Swag when we need its additional features?

yevgeny-shnaidman:
> @yevgeny-shnaidman I did not give it much thought - I discovered the Kubernetes library by accident and thought it would make sense to standardize on something coming from the Kubernetes org. Maybe we can use Swag when we need its additional features?

ok

yevgeny-shnaidman (approved):

PR #98, opened 2022-07-05
Add a new e2e workflow with a variable in the CR

qbarrand:
This new workflow tests that variable substitution introduced in #73 works as expected.


yevgeny-shnaidman:
what is the difference in the flow from the earlier version?

qbarrand:
@yevgeny-shnaidman This is actually a brand new flow. Currently, no flow is testing the variable substitution feature.

yevgeny-shnaidman (approved):

PR #104, opened 2022-07-08
Do not try to substitute Dockerfile variables

qbarrand:
PR #73 introduced variable substitution for kernel mappings. However it tried and replaced everyt string starting with `$` stored in the mapping struct, including in the Dockerfile. That substituted build arguments and environment variables in the Dockerfile with empty strings, breaking builds.
This PR changes the behaviour to only replace variables set in the `ContainerImage` field. If more kernel variables are desired at build time, we should add them as build arguments instead of replacing user-provided strings.
It also adds a check for build arguments in the `In-cluster build` e2e CI workflow.

### DIFF
--- a/internal/module/kernelmapper.go
+++ b/internal/module/kernelmapper.go
@@ -99,7 +99,7 @@ func (k *kernelMapper) PrepareKernelMapping(mapping *ootov1alpha1.KernelMapping,
 	substMapping := mapping.DeepCopy()
 	substMapping.ContainerImage = substContainerImage
 
-	return &substMapping, nil
+	return substMapping, nil
 }
 
 func (k *kernelMapper) prepareOSConfigList(osConfig NodeOSConfig) []string {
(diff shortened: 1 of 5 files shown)


internal/module/kernelmapper.go
```diff
@@ -88,19 +87,19 @@ func (k *kernelMapper) GetNodeOSConfig(node *v1.Node) *NodeOSConfig {
...
+	substContainerImage, err := parser.Parse(mapping.ContainerImage)
 	if err != nil {
-		return nil, fmt.Errorf("failed to substitute the os config into mapping: %w", err)
+		return nil, fmt.Errorf("failed to substitute the os config into ContainerImage field: %w", err)
 	}
-	substMapping := ootov1alpha1.KernelMapping{}
-	err = json.Unmarshal([]byte(parsedMapping), &substMapping)
-	return &substMapping, err
+
+	substMapping := *mapping
```
yevgeny-shnaidman, line 99: i suggests to use DeepCopy instead of pointer de-reference, more secure

yevgeny-shnaidman (approved):

PR #106, opened 2022-07-12
Add the PodNodeModule reconciler

qbarrand:
Add a separate reconciler that watches driver pods and labels the nodes they're running on with the `Module` name whenever they acquire the `Ready` condition.
The reconciler is triggered in the following situations:
- a new driver pod is created;
- a driver pod's `Ready` condition changes its value;
- a driver pod is deleted.

Do make sure we never miss any pod deletion and have node labels in sync with pods, we create pods with a finalizer that is removed by the new controller.

### DIFF
--- a/controllers/pod_node_module_reconciler.go
+++ b/controllers/pod_node_module_reconciler.go
@@ -1,0 +1,144 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/qbarrand/oot-operator/internal/constants"
+	"github.com/qbarrand/oot-operator/internal/daemonset"
+	"github.com/qbarrand/oot-operator/internal/filter"
+	v1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/kubectl/pkg/util/podutils"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+//+kubebuilder:rbac:groups="core",resources=pods,verbs=get;patch;list;watch
+//+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;watch
+
+type PodNodeModuleReconciler struct {
+	client client.Client
+}
+
+func NewPodNodeModuleReconciler(client client.Client) *PodNodeModuleReconciler {
+	return &PodNodeModuleReconciler{client: client}
+}
+
+func (pnmr *PodNodeModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := ctrl.LoggerFrom(ctx)
+
+	pod := v1.Pod{}
+	podNamespacedName := req.NamespacedName
+
+	if err := pnmr.client.Get(ctx, podNamespacedName, &pod); err != nil {
+		if k8serrors.IsNotFound(err) {
+			logger.Info("Pod not found")
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{}, fmt.Errorf("could not get pod %s: %v", podNamespacedName, err)
+	}
+
+	nodeName := pod.Spec.NodeName
+
+	moduleName, ok := pod.Labels[constants.ModuleNameLabel]
+	if !ok {
+		return ctrl.Result{}, fmt.Errorf("pod %s has no %q label", podNamespacedName, constants.ModuleNameLabel)
+	}
+
+	labelName := daemonset.GetDriverContainerNodeLabel(moduleName)
+
+	logger = logger.WithValues(
+		"node name", nodeName,
+		"module name", moduleName,
+		"label name", labelName,
+	)
+
+	if !podutils.IsPodReady(&pod) {
+		logger.Info("Unlabeling node")
+
+		if err := pnmr.deleteLabel(ctx, nodeName, labelName); err != nil {
+			return ctrl.Result{}, fmt.Errorf("could not unlabel node %s: %v", nodeName, err)
+		}
+
+		if !pod.DeletionTimestamp.IsZero() {
+			logger.Info("Pod deletion requested; removing finalizer")
+
+			if err := pnmr.deleteFinalizer(ctx, &pod); err != nil {
+				return ctrl.Result{}, fmt.Errorf("could not delete the pod finalizer: %v", err)
+			}
+		}
+
+		return ctrl.Result{}, nil
+	}
+
+	logger.Info("Labeling node")
+
+	if err := pnmr.addLabel(ctx, nodeName, labelName); err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not label node %s with %q: %v", nodeName, labelName, err)
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (pnmr *PodNodeModuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	p := predicate.And(
+		filter.PodReadinessChangedPredicate(
+			mgr.GetLogger().WithName("pod-readiness-changed"),
+		),
+		filter.HasLabel(constants.ModuleNameLabel),
+		filter.PodHasSpecNodeName(),
+	)
+
+	return ctrl.
+		NewControllerManagedBy(mgr).
+		Named("pod-node-module").
+		For(&v1.Pod{}).
+		WithEventFilter(p).
+		Complete(pnmr)
+}
+
+func (pnmr *PodNodeModuleReconciler) addLabel(ctx context.Context, nodeName, labelName string) error {
+	node := v1.Node{}
+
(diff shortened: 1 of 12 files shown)


controllers/pod_node_module_reconciler.go
```diff
@@ -0,0 +1,166 @@
...
+	deleting := !pod.DeletionTimestamp.IsZero()
+
+	logger = logger.WithValues(
+		"node name", nodeName,
+		"module name", moduleName,
+		"label name", labelName,
+		"deleting pod", deleting,
+	)
+
+	if !podutils.IsPodReady(&pod) || deleting {
```
yevgeny-shnaidman, line 66: i think you can remove deleting condition here: you are checking it anyway in line73, and pod cannot be both ready and deleting

controllers/pod_node_module_reconciler.go
```diff
@@ -0,0 +1,166 @@
...
+
+	delete(node.Labels, labelName)
+
+	return pnmr.client.Patch(ctx, &node, client.MergeFrom(nodeCopy))
+}
+
+// PodReadinessChangedPredicate returns a predicate for Update events that only returns true if the Ready condition
+// changed.
+func PodReadinessChangedPredicate(logger logr.Logger) predicate.Predicate {
+	return predicate.Funcs{
```
yevgeny-shnaidman, line 149: what do you think about moving all the predicates into the filter package? eventually this is what this package is for

internal/daemonset/daemonset.go
```diff
@@ -162,8 +166,28 @@ func (dc *daemonSetGenerator) SetDriverContainerAsDesired(ctx context.Context, d
 		},
 	}
 
-	return dc.constructDaemonSet(ctx, ds, &mod, mod.Spec.DriverContainer, "driver-container", image, standardLabels,
```
yevgeny-shnaidman, line 165: why not just use constructDaemonSet and pass it the finalizers?

controllers/pod_node_module_reconciler.go
```diff
@@ -0,0 +1,166 @@
...
+	pod := v1.Pod{}
+	podNamespacedName := req.NamespacedName
+
+	if err := pnmr.client.Get(ctx, podNamespacedName, &pod); err != nil {
+		if k8serrors.IsNotFound(err) {
+			logger.Info("Pod not found")
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{}, fmt.Errorf("could not get pod %s: %v", podNamespacedName, err)
```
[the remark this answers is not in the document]
qbarrand: My concern with `%w` is that you can do `errors.Is` at any point in the error return chain. It makes it possible for our code to check for error types that are implementation details in most cases and not part of the API.

controllers/pod_node_module_reconciler.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	nodeName := pod.Spec.NodeName
+
+	moduleName, ok := pod.Labels[constants.ModuleNameLabel]
```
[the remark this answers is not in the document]
qbarrand: I see the predicate as an optimization rather than a safeguard. If we remove that check, we might end up with a bad label. Would you still remove it?

controllers/pod_node_module_reconciler.go
qbarrand: Does deleting automatically remove the `Ready` condition? I could not find any documentation on this.

internal/daemonset/daemonset.go
qbarrand: That's yet another parameter for a limited value, IMHO. And I guess the usefulness of that function will even decrease when we refactor the CRD.

controllers/pod_node_module_reconciler.go
```diff
@@ -0,0 +1,166 @@
...
+			if err := pnmr.deleteFinalizer(ctx, &pod); err != nil {
+				return ctrl.Result{}, fmt.Errorf("could not delete the pod finalizer: %v", err)
+			}
+		}
+
+		return ctrl.Result{}, nil
+	} else {
+		logger.Info("Labeling node")
+
+		return ctrl.Result{}, pnmr.addLabel(ctx, nodeName, labelName)
```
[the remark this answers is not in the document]
qbarrand: Fixed.

controllers/pod_node_module_reconciler.go
```diff
@@ -0,0 +1,166 @@
...
+func PodReadinessChangedPredicate(logger logr.Logger) predicate.Predicate {
+	return predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldPod, ok := e.ObjectOld.(*v1.Pod)
+			if !ok {
+				logger.Info("Old object is not a pod", "object", e.ObjectOld)
+				return true
+			}
+
+			newPod, ok := e.ObjectNew.(*v1.Pod)
```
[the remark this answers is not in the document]
qbarrand: I think I am missing the comment above...

controllers/pod_node_module_reconciler.go
qbarrand: I am a bit torn on that one. It makes sense to have them all in the same place, but that predicate is quite specific and I think having it close to the reconciler makes the logic easier to understand.

controllers/pod_node_module_reconciler.go
```diff
@@ -0,0 +1,166 @@
...
+
+	return pnmr.client.Patch(ctx, &node, client.MergeFrom(nodeCopy))
+}
+
+// PodReadinessChangedPredicate returns a predicate for Update events that only returns true if the Ready condition
+// changed.
+func PodReadinessChangedPredicate(logger logr.Logger) predicate.Predicate {
+	return predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldPod, ok := e.ObjectOld.(*v1.Pod)
```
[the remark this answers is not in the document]
qbarrand: Ha, here is the comment above 🙂
We do `ctrl.NewControllerManagedBy(mgr).Named("pod-node-module").For(&v1.Pod{})` so the predicate should indeed only receive pods. But if for any reason that's not the case here, `oldPod` would be `nil`. Either we check if `oldPod` is `nil` or we use `ok`.

controllers/pod_node_module_reconciler.go
yevgeny-shnaidman: Delete is set by pod controller when the containers are deleted. Once the containers are deleted, pod cannot be in Ready state. But even if we have a race condition here, we will not fail, since we will catch it on the next reconciliation loop

controllers/pod_node_module_reconciler.go
yevgeny-shnaidman: I would prefer to have them all concentrated in one place. will allow optimizations of the code if need be, and makes it more clear to understand

internal/daemonset/daemonset.go
yevgeny-shnaidman: it just seems stange that we have it for device plugin and not for driver-container. essentially it is almost the same function flow

controllers/pod_node_module_reconciler.go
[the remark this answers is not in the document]
yevgeny-shnaidman: predicates are optimizations, but you can think of them as a code that is always run before calling reconcil, so we perform that check twice. But it is not a principal issue...

controllers/pod_node_module_reconciler.go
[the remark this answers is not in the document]
yevgeny-shnaidman: i simply would not change any of those conditions, they should all have been checked by the controller-runtime code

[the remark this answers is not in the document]
qbarrand: So you would not check if the type assertion was valid? The code will panic if for any reason `oldPod` or `newPod` are nil.

controllers/pod_node_module_reconciler.go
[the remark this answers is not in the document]
qbarrand: Created [MGMT-11158](https://issues.redhat.com/browse/MGMT-11158).

internal/daemonset/node_label.go
```diff
@@ -0,0 +1,7 @@
+package daemonset
+
+import "fmt"
+
+func NodeLabelName(moduleName string) string {
```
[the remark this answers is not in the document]
qbarrand: It should already have been removed, apologies.

internal/daemonset/daemonset.go
qbarrand: My plan is to remove that call for the device plugin creation as well. The two `SetAsDesired` functions do similar work indeed, but calling a common function with many parameters makes the code less straightforward IMHO.

controllers/pod_node_module_reconciler.go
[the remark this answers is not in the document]
yevgeny-shnaidman: ok, let's go with it

controllers/pod_node_module_reconciler.go
qbarrand: Indeed. Simplified the logic.

qbarrand:
Move of predicate aggregation to `internal/filter` to be tracked in [MGMT-11180](https://issues.redhat.com/browse/MGMT-11180).

internal/daemonset/daemonset.go
qbarrand: Removed the `constructDaemonSet` method.

yevgeny-shnaidman (approved):

PR #107, opened 2022-07-14
Adding additional metrics to the KMMO

yevgeny-shnaidman:
1) metric that reports the status of the stages of KMMO CR:
   build ( completed or not), driver container (completed or not), device plugin( completed or not)
2) metric that reports the targeted nodes ( by label seclector), the nodes that driver container
   will run on, based on label selector and kernel mapping) and the nodes that the driver container
   has currently been deployed to
3) changing the status update from Patch to Update

MGMT-10859

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -8,6 +8,10 @@ import (
 // When adding metric names, see https://prometheus.io/docs/practices/naming/#metric-names
 const (
 	existingKMMOModulesQuery = "kmmo_module_total"
+	completedKMMOStageQuery  = "kmmo_completed_stage"
+	BuildStage               = "build"
+	DriverContainerStage     = "driver-container"
+	DevicePluginStage        = "device-plugin"
 )
 
 //go:generate mockgen -source=metrics.go -package=metrics -destination=mock_metrics_api.go
@@ -16,10 +20,12 @@ const (
 type Metrics interface {
 	Register()
 	SetExistingKMMOModules(value int)
+	SetCompletedStage(kmmoName, kmmoNamespace, kernelVersion, stage string, completed bool)
 }
 
 type metrics struct {
-	kmmoResourcesNum prometheus.Gauge
+	kmmoResourcesNum   prometheus.Gauge
+	kmmoCompletedStage *prometheus.GaugeVec
 }
 
 func New() Metrics {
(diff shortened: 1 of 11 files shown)


internal/metrics/metrics.go
```diff
@@ -30,15 +39,47 @@ func New() Metrics {
 			Help: "Number of existing KMMO Modules",
 		},
 	)
-	return &metrics{kmmoResourcesNum: kmmoResourcesNum}
+	completedStages := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: completedKMMOStageQuery,
+			Help: "For a given kmmo,namespace, stage(device-plug, driver-container, build), 1 if the stage is completed, 0 if it is not.",
```
qbarrand, line 45: There can be many builds for a single `Module` CR. Should we also add e.g. the kernel version as a label?

internal/module/statusupdater.go
```diff
@@ -53,5 +57,42 @@ func (s *statusUpdater) UpdateModuleStatus(ctx context.Context, mod *ootov1alpha
...
+func (s *statusUpdater) updateMetrics(ctx context.Context, mod *ootov1alpha1.Module, targetedNodes, kernelMappingNodes []v1.Node) {
+	logger := log.FromContext(ctx)
+	s.metricsAPI.SetCompletedStage(mod.Name, mod.Namespace, metrics.DriverContainerStage, mod.Status.DriverContainer.DesiredNumber == mod.Status.DriverContainer.AvailableNumber)
+	if mod.Spec.DevicePlugin != nil {
+		s.metricsAPI.SetCompletedStage(mod.Name, mod.Namespace, metrics.DevicePluginStage, mod.Status.DevicePlugin.DesiredNumber == mod.Status.DevicePlugin.AvailableNumber)
+	}
+
+	deployedNodes, err := s.getCurrentlyDeployedNodes(ctx, mod)
+	if err != nil {
+		logger.Info("failed to get deployed nodes", "module", mod.Name, "error", err)
```
qbarrand, line 73: How about returning that error instead?

internal/metrics/metrics.go
```diff
@@ -30,15 +39,47 @@ func New() Metrics {
...
+		},
+		[]string{"kmmo", "namespace", "stage"},
+	)
+
+	nodesUtilization := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: nodesUtilizationQuery,
+			Help: "Nodes that driver-container/device-plugin shoud run on based on selector, nodes they will run based on kernel mappings, currently deploed nodes",
```
qbarrand, line 53: I think have highly variable labels will lead to huge cardinality. If we want to to expose a list of nodes to the user, maybe the `Module`'s status is a better place?

internal/module/statusupdater.go
yevgeny-shnaidman: not relevant anymore, as we won't supply node utilization metrics

internal/metrics/metrics.go
yevgeny-shnaidman: fixed

internal/metrics/metrics.go
yevgeny-shnaidman: agree, removing this metric, will open a PR for adding the info to the mod.Status

qbarrand (approved):

PR #111, opened 2022-07-18
Added best effort unit-tests for the `registry` package.

ybettan:

### DIFF
--- a/internal/registry/registry_test.go
+++ b/internal/registry/registry_test.go
@@ -1,0 +1,105 @@
+package registry
+
+import (
+	context "context"
+	"errors"
+
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	ootov1alpha1 "github.com/qbarrand/oot-operator/api/v1alpha1"
+	"github.com/qbarrand/oot-operator/internal/auth"
+	v1 "k8s.io/api/core/v1"
+)
+
+var _ = Describe("ImageExists", func() {
+
+	const (
+		psName       = "pull-push-secrert"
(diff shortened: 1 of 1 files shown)


internal/registry/registry.go
```diff
@@ -72,8 +72,8 @@ func (r *registry) ImageExists(ctx context.Context, image string, po ootov1alpha
 	return true, nil
 }
 
-func (r *registry) GetLayersDigests(ctx context.Context, image string) ([]string, *RepoPullConfig, error) {
-	pullConfig, err := r.getPullOptions(ctx, image, nil, nil, "")
+func (r *registry) GetLayersDigests(ctx context.Context, image string, ps *corev1.LocalObjectReference, psNamespace string) ([]string, *RepoPullConfig, error) {
+	pullConfig, err := r.getPullOptions(ctx, image, nil, ps, psNamespace)
```
qbarrand, line 76: We inject an authenticator when building the registry. Then `getPullOptions` will add the keychain to the options it returns.
I don't think we need to get `ps` and `psNamespace` as parameters to this function then.

ybettan: > Then getPullOptions will add the keychain to the options it returns

In order to do so it need to access the secret in the cluster.

> I don't think we need to get ps and psNamespace as parameters to this function then.

I don't see how you can do it otherwise, how will you get the user pull-secret if you don't know what is the secret name?

ybettan: As agreed.

I will remove that change from this PR and make this PR a unit-test PR only.

Then we will have a separate PR with making the `registry` package unaware of the underlying secret usage so we can easily extend it to support `volt` for example in the future.

PR #113, opened 2022-07-19
Injecting the registry auth getter interface to the registry's methods.

ybettan:
By creating an auth getter on every reconcile iteration, and injecting
it into the registry's methods we increase the code flexibility ease the
task of adding additional auth getter in the future.


internal/auth/auth.go
```diff
@@ -13,29 +14,31 @@ import (
 
 //go:generate mockgen -source=auth.go -package=auth -destination=mock_auth.go
 
-type RegistryAuth interface {
-	GetKeyChainFromSecret(ctx context.Context, secretName, secretNamespace string) (authn.Keychain, error)
+type RegistryAuthGetter interface {
```
qbarrand, line 17: How about `RegistryAuthenticator`?

internal/auth/auth.go
```diff
@@ -13,29 +14,31 @@ import (
...
+	GetKeyChain(ctx context.Context) (authn.Keychain, error)
 }
 
-type registryAuth struct {
-	client client.Client
+type registrySecretAuthGetter struct {
+	client         client.Client
+	namespacedName *types.NamespacedName
```
qbarrand, line 23: I don't think we need to store a pointer here, how about `namespacedName` being a `types.NamespacedName`?

qbarrand (commented):
I think it looks great. If we ever need to pass more objects to the individual `Registry` methods, we can always implement the [Options pattern](https://www.sohamkamani.com/golang/options-pattern/) like `crane` does. But for now, this is more than enough.

internal/auth/auth.go
ybettan: I have though about it, but it doesn't do any authentication, it only get the keychain from somewhere (in this case from a secret in the cluster), therefore, I think that `RegistryAuthGetter` is a better description of what the interface actually does.

internal/auth/auth.go
ybettan: > I don't think we need to store a pointer here

Why not? it is making it easier for the `if` blocks that only need to check `if registryAuthGetter.namespacedName != nil`

internal/auth/auth.go
qbarrand: Fair enough.

internal/auth/auth.go
ybettan: Fixed.

internal/build/job/manager.go
```diff
@@ -65,7 +67,19 @@ func (jbm *jobManager) Sync(ctx context.Context, mod ootov1alpha1.Module, m ooto
...
 	buildConfig := jbm.helper.GetRelevantBuild(mod, m)
 
-	imageAvailable, err := jbm.registry.ImageExists(ctx, m.ContainerImage, buildConfig.Pull, mod.Spec.ImagePullSecret, mod.Namespace)
+	var (
+		namespacedName     types.NamespacedName
+		registryAuthGetter auth.RegistryAuthGetter
+	)
+
+	if mod.Spec.ImagePullSecret != nil {
+		namespacedName = types.NamespacedName{
```
qbarrand, line 76:
```suggestion
	var registryAuthGetter auth.RegistryAuthGetter

	if mod.Spec.ImagePullSecret != nil {
		namespacedName := types.NamespacedName{
```

internal/auth/auth.go
```diff
@@ -13,29 +13,27 @@ import (
...
 
 	secret := v1.Secret{}
-	secretNamespacedName := types.NamespacedName{
-		Name:      secretName,
-		Namespace: secretNamespace,
-	}
-	if err := a.client.Get(ctx, secretNamespacedName, &secret); err != nil {
-		return nil, fmt.Errorf("cannot find secret %s: %w", secretNamespacedName.String(), err)
+	if err := rsag.client.Get(ctx, rsag.namespacedName, &secret); err != nil {
+		return nil, fmt.Errorf("cannot find secret %s: %w", rsag.namespacedName.String(), err)
```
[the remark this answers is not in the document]
ybettan: Removed.

internal/build/job/manager_test.go
```diff
@@ -163,7 +166,40 @@ var _ = Describe("JobManager", func() {
...
+			gomock.InOrder(
+				helper.EXPECT().GetRelevantBuild(mod, km).Return(km.Build),
+				registry.EXPECT().ImageExists(ctx, imageName, po, gomock.Any()),
+				maker.EXPECT().MakeJob(mod, km.Build, kernelVersion, km.ContainerImage).Return(&j, nil),
+			)
+
+			gomock.InOrder(
+				clnt.EXPECT().List(ctx, gomock.Any(), gomock.Any(), gomock.Any()),
+				clnt.EXPECT().Create(ctx, &j),
+			)
```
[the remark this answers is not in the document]
ybettan: Done.

qbarrand (approved):

PR #116, opened 2022-07-20
Garbage collector should skip handling/checking device-plugin daemonset

yevgeny-shnaidman:
MGMT-11228

### DIFF
--- a/internal/daemonset/daemonset.go
+++ b/internal/daemonset/daemonset.go
@@ -54,7 +54,7 @@ func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, existingDS map
 	deleted := make([]string, 0)
 
 	for kernelVersion, ds := range existingDS {
-		if !validKernels.Has(kernelVersion) {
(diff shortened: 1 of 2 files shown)


internal/daemonset/daemonset.go
```diff
@@ -54,7 +54,7 @@ func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, existingDS map
 	deleted := make([]string, 0)
 
 	for kernelVersion, ds := range existingDS {
-		if !validKernels.Has(kernelVersion) {
+		if !dc.isDevicePluginDaemonSet(ds) && !validKernels.Has(kernelVersion) {
```
qbarrand, line 57: I think we could instead make `existingDS` contain only driver DaemonSets. WDYT?

yevgeny-shnaidman: it used to previously, but i changed it to contain also device plugin, since it allows us to use this info in the status updater and metrics, without additional queries

qbarrand: We agree to refactor if this ever leads to many ifs in the codebase.

PR #117, opened 2022-07-20
Always reconcile pods being deleted

qbarrand:
Because we only reconciled pod that had their `Ready` condition changed, we would never remove the finalizer on pods that never become `Ready`.

This PR also makes the CI build container images for each PR for easier testing.


enriquebelarte (commented):
Changes tested and working. Thanks!

enriquebelarte (approved):

PR #121, opened 2022-07-21
Node that the device plugin pod is running on

yevgeny-shnaidman:
should be labeled accordingly

MGMT-11208


qbarrand (approved):
We should work on an e2e test for the device plugin feature at some point.

PR #123, opened 2022-07-21
MGMT-10525: Build must-gather image to collect KMMO information

erusso7:
The new script will collect information from some resources to help us and our customers to debug and troubleshoot. 
It can be parametrized by setting the following environment variables:
- `NS` for the namespace where the operator is running. If it's not specified, it will try to identify it automatically. 
- `OUTPUT_DIR` for the destination of the collected information. 

Gathered information:
- Inspection (`oc adm inspect`) of all the resources within the `NS` namespace.
- Logs of the `controller-manager` 
- Inspection of the `crd/modules.ooto...` in all the namespaces because it's cluster scoped resource. 
- Inspection of `Daemonsets` in all the namespaces where they exist. 
- Inspection of `Jobs` in all the namespaces where they were created.


Dockerfile.must-gather
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2021 Red Hat, Inc.
+# Copyright Contributors to the Open Cluster Management project
```
qbarrand, line 2: Should we remove that line?

Makefile
```diff
@@ -35,6 +35,9 @@ IMAGE_TAG_BASE ?= ghcr.io/qbarrand/oot-operator
 # You can use it as an arg. (E.g make bundle-build BUNDLE_IMG=<some-registry>/<project-name-bundle>:<tag>)
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 
+# GATHER_IMG define the must-gather image used to get all the information for debugging purpose.
+GATHER_IMG ?= quay.io/rh_ee_erusso/kmmo-must-gather
```
qbarrand, line 39: Let's instead try to have it built by the GitHub CI and store it there.
Could you have a look at `.github/workflows/container-image.yml` and replicate what we do for other images?

must-gather/gather
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
+TIMESTAMP=$(date +%Y%m%d_%H%M%S)
+OUTPUT_DIR="${OUTPUT_DIR:-must-gather/${TIMESTAMP}}"
+
+DEFAULT_NS=$(oc get ns -o name | grep "oot" | cut -d"/" -f 2)
+NS="${NS:-$DEFAULT_NS}"
```
qbarrand, line 7: I think we could just do
```suggestion
NS="${NS:-oot-operator-system}"
```
It will be easier to find when we replace all `oot` references to `kmm`.

must-gather/gather
```diff
@@ -0,0 +1,37 @@
...
+  echo "Collecting information for ${resource}"
+  describeFile="${OUTPUT_DIR}/${type}_${name}.yaml"
+  oc -n "$NS" describe "$resource" > "$describeFile"
+done
+
+echo "Collecting logs from 'deployment/oot-operator-controller-manager'"
+oc -n "$NS" logs "deployment/oot-operator-controller-manager" > "${OUTPUT_DIR}/oot-operator-controller-manager.log"
+
+echo "Inspecting modules.ooto.sigs.k8s.io in all the namespaces"
+oc adm inspect -A crd/modules.ooto.sigs.k8s.io  --dest-dir="$OUTPUT_DIR/inspect"
```
qbarrand, line 27: This just inspects the `Module` CRD. We also want all `Modules` in all namespaces:
```suggestion
echo "Inspecting the modules.ooto.sigs.k8s.io CRD"
oc adm inspect crd/modules.ooto.sigs.k8s.io  --dest-dir="$OUTPUT_DIR/inspect"

echo "Inspecting all Modules in all namespaces"
oc adm inspect -A modules.ooto.sigs.k8s.io --dest-dir="$OUTPUT_DIR/inspect"
```

must-gather/gather
```diff
@@ -0,0 +1,37 @@
...
+
+for ns in $(oc get daemonset -A -l oot.node.kubernetes.io/module.name --no-headers -o custom-columns="NS:.metadata.namespace"); do
+  echo "Inspecting DaemonSet in '${ns}' namespace"
```
qbarrand, line 30: Does this work? These are single quotes around `${ns}`, is the variable substitution still happening?

erusso7: Yes, because they are surrounded by double quotes. The single quotes are part of the content of the output.

must-gather/gather
erusso7: Yes, this is what I did in the previous one, I will revert it.

qbarrand (approved):

PR #125, opened 2022-07-25
Add a Preflight/Prebuild CRD definition

yevgeny-shnaidman:
MGMT-10680


api/v1alpha1/preflightvalidation_types.go
```diff
@@ -0,0 +1,116 @@
...
+// PreflightValidationSpec describes the desired state of the resource, such as the OCP image
+// that SR CRs need to be verified against and the debug configuration of the logs
+// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+// +kubebuilder:validation:Required
+type PreflightValidationSpec struct {
+	// Debug enables additional logging.
+	// +kubebuilder:validation:Optional
+	Debug bool `json:"debug"`
+
+	// KernelImage describes the kernel image that all KMMO CRs need to be checked against.
```
qbarrand, line 39:
```suggestion
	// KernelImage describes the kernel image that all Modules need to be checked against.
```

api/v1alpha1/preflightvalidation_types.go
```diff
@@ -0,0 +1,116 @@
...
+
+const (
+	VerificationTrue    string = "True"
+	VerificationFalse   string = "False"
+	VerificationStageImage   string = "Image"
+	VerificationStageBuild string = "Build"
+)
+
+// PreflightValidationSpec describes the desired state of the resource, such as the OCP image
+// that SR CRs need to be verified against and the debug configuration of the logs
```
qbarrand, line 31:
```suggestion
// PreflightValidationSpec describes the desired state of the resource, such as the kernel version
// that Module CRs need to be verified against as well as the debug configuration of the logs.
```

api/v1alpha1/preflightvalidation_types.go
```diff
@@ -0,0 +1,116 @@
...
+// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+// +kubebuilder:validation:Required
+type PreflightValidationSpec struct {
+	// Debug enables additional logging.
+	// +kubebuilder:validation:Optional
+	Debug bool `json:"debug"`
+
+	// KernelImage describes the kernel image that all KMMO CRs need to be checked against.
+	// +kubebuilder:validation:Required
+	KernelImage string `json:"kernelImage"`
```
qbarrand, line 41:
```suggestion
	KernelVersion string `json:"kernelVersion"`
```

api/v1alpha1/preflightvalidation_types.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// PreflightValidationSpec describes the desired state of the resource, such as the OCP image
+// that SR CRs need to be verified against and the debug configuration of the logs
+// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+// +kubebuilder:validation:Required
+type PreflightValidationSpec struct {
+	// Debug enables additional logging.
+	// +kubebuilder:validation:Optional
+	Debug bool `json:"debug"`
```
qbarrand, line 37: I understand we will be exposing the validation status for each `Module`. Do we really need that boolean then? What will it govern? I think might be confusing to have it trigger more logs on the standard output - log level should be configurable in one place only.

api/v1alpha1/preflightvalidation_types.go
```diff
@@ -0,0 +1,116 @@
...
+	VerificationFalse   string = "False"
+	VerificationStageImage   string = "Image"
+	VerificationStageBuild string = "Build"
+)
+
+// PreflightValidationSpec describes the desired state of the resource, such as the OCP image
+// that SR CRs need to be verified against and the debug configuration of the logs
+// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+// +kubebuilder:validation:Required
+type PreflightValidationSpec struct {
```
qbarrand, line 34: Shall we also give this spec a list of labels like a node would have? Then it really can behave like a node and be targeted by some `Modules` only like a node would.

api/v1alpha1/preflightvalidation_types.go
yevgeny-shnaidman: fixed

api/v1alpha1/preflightvalidation_types.go
yevgeny-shnaidman: fixed

api/v1alpha1/preflightvalidation_types.go
yevgeny-shnaidman: fixed

api/v1alpha1/preflightvalidation_types.go
yevgeny-shnaidman: Not sure i get the idea... Preflight verifies Modules and builds images , if need be, it does not deploy anything permanent onto the nodes

api/v1alpha1/preflightvalidation_types.go
yevgeny-shnaidman: The idea was to enable more logs for this specific Preflight run. Since each Preflight run is eventually final, if there is any issues that we want to debug with logs, we can just apply a new Preflight CR with debug enabled

qbarrand: > each Preflight run is eventually final

I think it should rather run continuously, each time a `Module` changes. WDYT?

api/v1alpha1/preflightvalidation_types.go
qbarrand: Right now, the `PreflightValidation` will verify all modules with a given kernel version, including `Modules` that may not have a valid mapping.
If we added labels to that object, we could restrict that validation to some Modules only (through their `selector` field).

api/v1alpha1/preflightvalidation_types.go
yevgeny-shnaidman: I am not sure about it ( from user's POV). Preflight CR should be applied once a week at best, probably even less frequently. I think once it has finished, user will simply delete, and create a new one next time, instead of just keeping it, and remembering that it is there and then updating it. Also, what happens if a Module CR changes, but the preflight has already been fully reconciled. Do we start Preflight reconciliation again? But user will not see the result, since he is not monitoring the Preflight

yevgeny-shnaidman: fixed

api/v1alpha1/preflightvalidation_types.go
yevgeny-shnaidman: as  discussed, we won't use labels in the Preflight

qbarrand (approved):

qbarrand (approved):

PR #127, opened 2022-07-27
New modprobe-oriented CRD

qbarrand:
This PR makes changes to the CRD and crafts the `modprobe` command for the driver container.


qbarrand:
@chr15p this PR includes the [node selector for builds](https://github.com/qbarrand/oot-operator/pull/127/files#diff-3047cc52e57eb78321dde3abdbad760dc862ba7d07b6130015c2346ee29e74c0R59) which is useful for multi-arch clusters.

ybettan:
Don't forget to squash before merging :)

api/v1alpha1/module_types.go
```diff
@@ -92,42 +93,112 @@ type KernelMapping struct {
...
-	Build *Build `json:"build"`
+	Args *ModprobeArgs `json:"args,omitempty"`
 
+	// If RawArgs are specified, they are passed straight to the modprobe binary; all other properties in this
+	// object are ignored.
 	// +optional
-	// AdditionalVolumes is a list of volumes that will be attached to the DriverContainer / DevicePlugin pod,
-	// in addition to the default ones.
-	AdditionalVolumes []v1.Volume `json:"additionalVolumes"`
+	RawArgs *ModprobeArgs `json:"rawArgs,omitempty"`
```
ybettan, line 124: Can you please show an `modprobe` command example and explain what is the difference between `Args` and `RawArgs`?

api/v1alpha1/module_types.go
```diff
@@ -92,42 +93,112 @@ type KernelMapping struct {
...
+}
 
+type DriverContainerContainerSpec struct {
```
ybettan, line 127: I don't think there is a clear, logical, distinction between `DriverContainer` and `DriverContainerSpec`? What is the purpose of this separation?

For example I don't see why `KernelMapping` is a better fit in `DriverContainerContainerSpec` rather than in `DriverContainerSpec` or why `ImageRepoSecret` isn't a part of `DriverContainerContainerSpec` but `ImagePullPolicy` is.

WDYT about having a single object `DriverContainerSpec` for all those fields?

api/v1alpha1/module_types.go
```diff
@@ -92,42 +93,112 @@ type KernelMapping struct {
...
+	Container DevicePluginContainerSpec `json:"container"`
+
+	ImagePullSecrets []v1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
 
 	// +optional
 	// ServiceAccountName is the name of the ServiceAccount to use to run this pod.
 	// More info: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 
+	Volumes []v1.Volume `json:"volumes,omitempty"`
```
ybettan, line 186: Although in k8s there is a reason for having `VolumeMounts` and `Volume` in 2 different objects (the reason is that a `Volume` is added at the pod level but for each container it need to be mapped to a `VolumeMount` location), I don't see the reason for it here.

As I wrote for `DriverContainerSpec`, here as well I don't see a logical separation between `DevicePluginSpec` and `DevicePluginContainerSpec`.

What is the reason for those separations?

internal/daemonset/daemonset.go
```diff
@@ -310,3 +322,44 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
 
 	return labels
 }
+
+func MakeLoadCommand(spec ootov1alpha1.ModprobeSpec) []string {
+	loadCommand := []string{"modprobe"}
+
+	if ra := spec.RawArgs; ra != nil && len(ra.Load) > 0 {
```
ybettan, line 329: Why do we need to check that `len(ra.Load)>0` if it is required in the spec that `minItems=1`?

internal/daemonset/daemonset.go
```diff
@@ -310,3 +322,44 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
...
+		return append(loadCommand, ra.Load...)
+	}
+
+	if a := spec.Args; a != nil && len(a.Load) > 0 {
```
ybettan, line 333: Why do we need to check that len(ra.Load)>0 if it is required in the spec that minItems=1?

internal/daemonset/daemonset.go
```diff
@@ -310,3 +322,44 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
...
+		loadCommand = append(loadCommand, a.Load...)
+	} else {
+		loadCommand = append(loadCommand, "-v")
```
ybettan, line 336: Why do we use `-v` only in that case?

internal/daemonset/daemonset.go
```diff
@@ -310,3 +322,44 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
...
+	}
+
+	if dirName := spec.DirName; dirName != "" {
+		loadCommand = append(loadCommand, "-d", dirName)
```
ybettan, line 340: Where are we setting the default value `/opt` if the user didn't specified any?

internal/daemonset/daemonset.go
```diff
@@ -310,3 +322,44 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
...
+	}
+
+	loadCommand = append(loadCommand, spec.ModuleName)
+	return append(loadCommand, spec.Parameters...)
+}
+
+func MakeUnloadCommand(spec ootov1alpha1.ModprobeSpec) []string {
+	unloadCommand := []string{"modprobe"}
```
ybettan, line 348:
```suggestion
	unloadCommand := []string{"modprobe", "-rv"}
```

internal/daemonset/daemonset.go
```diff
@@ -310,3 +322,44 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
...
+
+	if ra := spec.RawArgs; ra != nil && len(ra.Load) > 0 {
+		return append(unloadCommand, ra.Load...)
+	}
+
+	if a := spec.Args; a != nil && len(a.Unload) > 0 {
+		unloadCommand = append(unloadCommand, a.Unload...)
+	} else {
+		unloadCommand = append(unloadCommand, "-rv")
```
ybettan, line 357: I think we should always add `"-rv"`. Are we expecting the user to add the `-r` on each module?

internal/daemonset/daemonset.go
```diff
@@ -310,3 +322,44 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
 
 	return labels
 }
+
+func MakeLoadCommand(spec ootov1alpha1.ModprobeSpec) []string {
+	loadCommand := []string{"modprobe"}
```
ybettan, line 327:
```suggestion
	loadCommand := []string{"modprobe", "-v"}
```

internal/daemonset/daemonset.go
```diff
@@ -310,3 +322,44 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
...
+	}
+
+	if a := spec.Args; a != nil && len(a.Unload) > 0 {
+		unloadCommand = append(unloadCommand, a.Unload...)
+	} else {
+		unloadCommand = append(unloadCommand, "-rv")
+	}
+
+	if dirName := spec.DirName; dirName != "" {
+		unloadCommand = append(unloadCommand, "-d", dirName)
```
ybettan, line 361: What about handling the default value `/opt`?

api/v1alpha1/module_types.go
```diff
@@ -50,13 +50,14 @@ type PushOptions struct {
...
 
 type Build struct {
-
 	// +optional
 	// BuildArgs is an array of build variables that are provided to the image building backend.
 	BuildArgs []BuildArg `json:"buildArgs"`
 
 	Dockerfile string `json:"dockerfile"`
 
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
yevgeny-shnaidman, line 59: why do we need NodeSelector in Build? Why should user care where on which node the build is running?

api/v1alpha1/module_types.go
```diff
@@ -92,42 +93,112 @@ type KernelMapping struct {
...
+
+	// VolumeMounts is a list of volume mounts that are appended to the default ones.
+	// +optional
+	VolumeMounts []v1.VolumeMount `json:"volumeMounts,omitempty"`
+}
+
+type DevicePluginSpec struct {
+	Container DevicePluginContainerSpec `json:"container"`
+
+	ImagePullSecrets []v1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
```
yevgeny-shnaidman, line 179: Why is ImagePullSecret specific to the DevicePlugin and not the whole Module?

api/v1alpha1/module_types.go
chr15p: because if you have a hybrid cluster with ARM and x86 nodes you need to compile your driver on the same arch as you're deploying onto.

api/v1alpha1/module_types.go
ybettan: I agree that it will be clearer if we only have `ImageRepoSecret` at the module level to contain all the auth for all image registries in a single dockerconfig.json file.

Isn't what we have agreed on?

api/v1alpha1/module_types.go
qbarrand: Your typical `modprobe` command to load a module is `modprobe -v mymod`.
- `Args` lets you specify your own arguments instead of just `-v`: `modprobe ${Args} mymod`
- `RawArgs` lets you specify the entire command, in case you want to so something really custom with the arguments that KMMO does not support: `modprobe ${RawArgs}`

Specify `RawArgs` makes KMMO skip the generation of the CLI - it just appends whatever you provided to `modprobe` and runs that.

api/v1alpha1/module_types.go
qbarrand: I agree it might be a bit confusing.
- `DriverContainerSpec` is `.spec.driverContainer`. Contrary to what the name says, it contains all properties related to the driver **pod**. Right now that pod only contains the container that runs `modprobe` cont, but that might change in the future (maybe?)
- `DriverContainerContainerSpec` is `.spec.driverContainer.container`, which contains properties related to the first container in the driver pod (that runs `modprobe`).

The kernel mappings are in the container properties because they govern the `image` field in there. `ImageRepoSecret` is in `.spec.driverContainer` because the corresponding `.spec.imagePullSecrets` field is part of the [`PodSpec`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podspec-v1-core).

Maybe we should find a new name for the driver pod, because calling it DriverContainer is indeed confusing.

api/v1alpha1/module_types.go
qbarrand: Same as above. `DevicePluginSpec` contains all properties related to the pod, while `DevicePluginContainerSpec` contains all properties related to the first container in that pod.

internal/daemonset/daemonset.go
qbarrand: [In the CRD](https://github.com/qbarrand/oot-operator/pull/127/files/d3a3896d24aef1bd487e2008dbb62e9dee21151c#diff-3047cc52e57eb78321dde3abdbad760dc862ba7d07b6130015c2346ee29e74c0R114).

internal/daemonset/daemonset.go
qbarrand: This would not work well with `Args`.

internal/daemonset/daemonset.go
qbarrand: It is in `.spec.DirName` as definedin the CRD.

internal/daemonset/daemonset.go
qbarrand: We are expecting users not to use `Args`. But if they do, they should be able to do everything they want - including overwriting the default `-rv`.

api/v1alpha1/module_types.go
ybettan: Thanks. It make more sense know.

I do think we should add those examples you put in this comment inside the fields description for clarification though.

api/v1alpha1/module_types.go
ybettan: Technically, a secret will be referenced in the `volume` inside the pod's spec, and then mounted into a filesystem path at the container's spec.

I think it will be more clear to have a single object `spec.driverContainer` to hold all those fields without a separation between pod and container's specs don't you think so?

api/v1alpha1/module_types.go
ybettan: Will you suggest having them all under a single object `DevicePluginSpec`. I think it will be less confusing. It feels like we are enforcing our CRD to get the pod/container structure but those are different objects eventually.

internal/daemonset/daemonset.go
```diff
@@ -310,3 +322,44 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
...
+	loadCommand = append(loadCommand, spec.ModuleName)
+	return append(loadCommand, spec.Parameters...)
+}
+
+func MakeUnloadCommand(spec ootov1alpha1.ModprobeSpec) []string {
+	unloadCommand := []string{"modprobe"}
+
+	if ra := spec.RawArgs; ra != nil && len(ra.Load) > 0 {
+		return append(unloadCommand, ra.Load...)
+	}
```
ybettan, line 352: Is there a way to tell the CRD to respect mutual exclusion between using `paramaters/args` and using `rawArgs`? if not it means that if a user set them both we will ignore `parameters/args`.

internal/daemonset/daemonset.go
ybettan: Now that I understand better the difference between `args`, `rawArgs` and `parameters` it make more sense.

internal/daemonset/daemonset.go
qbarrand: I think that is okay if it is documented in the CRD. Do you agree?

api/v1alpha1/module_types.go
yevgeny-shnaidman: I think that in case of multi-arch architecture, KMMO should be the one that decides where to run the Build job and will know how to place it correctly.

internal/daemonset/daemonset.go
ybettan: If there is a possibility to enforce it by kubebuilder or something I think it is better, if not, let's just document it.

qbarrand: I don't think OpenAPI constraints are that flexible. But this check could be made by the [validating webhook](https://issues.redhat.com/browse/MGMT-10520) we are working on that verifies the regexes in the mappings.

internal/daemonset/daemonset.go
qbarrand: This would not work well with `Args`.

api/v1alpha1/module_types.go
qbarrand: > It feels like we are enforcing our CRD to get the pod/container structure

Yes. If for any reason we need to add properties for a second container in the pod, we just need to add another sub-object - there is a clear separation between what belongs to the pod and what belongs to the container.

> but those are different objects eventually.

Not sure I understand your point here.

api/v1alpha1/module_types.go
qbarrand: How do we determine the target architecture of a given image?

api/v1alpha1/module_types.go
qbarrand: Here are the current comments:
```go
// Args is an optional list of arguments to be passed to modprobe before the name of the kernel module.

// If RawArgs are specified, they are passed straight to the modprobe binary; all other properties in this
// object are ignored.
```
Do you think they are not clear enough?

api/v1alpha1/module_types.go
qbarrand: Switched to this.

api/v1alpha1/module_types.go
qbarrand: Now using the `Module`'s top-level `selector`.

api/v1alpha1/module_types.go
```diff
@@ -92,42 +91,112 @@ type KernelMapping struct {
...
+	// Unload is an optional list of arguments to be used when unloading the kernel module.
+	// +kubebuilder:validation:MinItems=1
+	Unload []string `json:"unload,omitempty"`
+}
+
+type ModprobeSpec struct {
+	// ModuleName is the name of the Module to be loaded.
+	ModuleName string `json:"moduleName"`
 
+	// Parameters is an optional list of kernel module parameters to be provided to modprobe.
```
ybettan, line 108:
```suggestion
	// Parameters is an optional list of kernel module parameters to be provided to modprobe.
	// Example: `modprobe -v mymod ${Parameters}`
```

I think this example worth more than any verbal explanation (especially for those who aren't familiar with the `modprobe` command syntax.

api/v1alpha1/module_types.go
```diff
@@ -92,42 +91,112 @@ type KernelMapping struct {
...
+	Parameters []string `json:"parameters,omitempty"`
+
+	// DirName is the root directory for modules.
+	// +kubebuilder:default=/opt
+	DirName string `json:"dirName,omitempty"`
+
+	// Args is an optional list of arguments to be passed to modprobe before the name of the kernel module.
```
ybettan, line 115:
```suggestion
	// Args is an optional list of arguments to be passed to modprobe before the name of the kernel module.
	// Example: `modprobe ${Args} mymod ...`
```

I think this example worth more than any verbal explanation (especially for those who aren't familiar with the modprobe command syntax.

api/v1alpha1/module_types.go
```diff
@@ -92,42 +91,112 @@ type KernelMapping struct {
...
 	// +optional
-	Build *Build `json:"build"`
+	Args *ModprobeArgs `json:"args,omitempty"`
 
+	// If RawArgs are specified, they are passed straight to the modprobe binary; all other properties in this
+	// object are ignored.
```
ybettan, line 120:
```suggestion
	// If RawArgs are specified, they are passed straight to the modprobe binary; all other properties in this
	// object are ignored.
	// Example: `modprobe ${RawArgs}
`
```

I think this example worth more than any verbal explanation (especially for those who aren't familiar with the modprobe command syntax.

api/v1alpha1/module_types.go
ybettan: I think the examples are super valuable here, I have commented on each one of them individually.
Resolving this thread.

api/v1alpha1/module_types.go
ybettan: So are we keeping the `NodeSelector`? The other (and more automated way) will be to:
* use podman to build and release multi-arch images, then, it doesn't matter where the build is scheduled, the correct manifest of the image will be chosen.

This will require additional update to DTK's documentation, and need to be validated.

internal/daemonset/daemonset.go
qbarrand: We might have only specified `unload` arguments and let `load` empty. That is valid from the CRD's perspective: `minItems=1` means "at least one item when the list is not nil".

internal/daemonset/daemonset.go
qbarrand: Same reason as above.

api/v1alpha1/module_types.go
qbarrand: PTAL.

api/v1alpha1/module_types.go
qbarrand: PTAL.

api/v1alpha1/module_types.go
ybettan: lgtm

api/v1alpha1/module_types.go
ybettan: lgtm

api/v1alpha1/module_types.go
ybettan: lgtm

ybettan (approved):

PR #130, opened 2022-08-02
Move StatusUpdater to a separate package and expand interface

yevgeny-shnaidman:
StatusUpdater interface will now handle also status updates for
Preflight CRs and, in the future, will also handle the Hub and
Spoke CR status update. This PR moves it into a separate function
and adds Preflight functions to the interface

MGMT-11431


internal/statusupdater/statusupdater.go
```diff
@@ -1,24 +1,29 @@
...
-		targetedNodes []v1.Node,
-		dsByKernelVersion map[string]*appsv1.DaemonSet) error
+	UpdateModuleStatus(ctx context.Context, mod *ootov1alpha1.Module, kernelMappingNodes []v1.Node,
+		targetedNodes []v1.Node, dsByKernelVersion map[string]*appsv1.DaemonSet) error
+	PresetPreflightVerificationStatus(ctx context.Context, pv *ootov1alpha1.PreflightValidation,
+		status *ootov1alpha1.CRStatus) error
+	SetPreflightVerificationStatus(ctx context.Context, preflight *ootov1alpha1.PreflightValidation, status *ootov1alpha1.CRStatus,
+		verificationStatus string, message string) error
+	SetPreflightVerificationStage(ctx context.Context, preflight *ootov1alpha1.PreflightValidation,
+		status *ootov1alpha1.CRStatus, stage string) error
```
ybettan, line 26: How about having the `StatusUpdater` interface to consist of 2 (3 in the future) private embedded interfaces for clear separation of concern?
Something like:
```
type StatusUpdated interface {
    moduleStatusUpdater
    preflightStatusUpdater
}
```

If you don't like it, how about at least have a better prefix naming?
```
ModuleUpdateStatus(...) error
PreflightGetVerificationStatus(...) error
PreflightSetVerificationStatus(...) error
PreflightSetVerificationStage(...) 
```
?

internal/statusupdater/statusupdater.go
```diff
@@ -64,6 +69,34 @@ func (s *statusUpdater) UpdateModuleStatus(ctx context.Context,
 	return s.client.Status().Update(ctx, mod)
 }
 
+func (s *statusUpdater) PresetPreflightVerificationStatus(ctx context.Context,
```
ybettan, line 72: I don't see any usage of that method so I am guessing one will call it with?
```
PresetPreflightVerificationStatus(ctx, &pv, &pv.status)
```

If this is the case, why do we need both `pv` and `status` as arguments?

Same applied for the other methods in this package.

internal/statusupdater/statusupdater.go
yevgeny-shnaidman: I don't see a reason to define additional interface that we are not going to use anyway. I will change the prefixes as suggested

internal/statusupdater/statusupdater.go
yevgeny-shnaidman: This function will be called from the preflight package ( PRs for that one will follow).
pv is the whole PreflightValidation, and CRStatus(ps) is just on entry in the array of CRStatuses that comprise the PreflightValidationStatus (see api/v1alpha1/preflightvalidation_types.go ) 
The calling method already obtains the needed CRStatus entry , since it needs it for its own logic, so i don't see the reason not to pass it as an paramter to the functions

internal/statusupdater/statusupdater.go
```diff
@@ -1,24 +1,29 @@
...
-		targetedNodes []v1.Node,
-		dsByKernelVersion map[string]*appsv1.DaemonSet) error
+	ModuleUpdateStatus(ctx context.Context, mod *ootov1alpha1.Module, kernelMappingNodes []v1.Node,
+		targetedNodes []v1.Node, dsByKernelVersion map[string]*appsv1.DaemonSet) error
+	PreflightPresetVerificationStatus(ctx context.Context, pv *ootov1alpha1.PreflightValidation,
+		status *ootov1alpha1.CRStatus) error
+	PreflightSetVerificationStatus(ctx context.Context, preflight *ootov1alpha1.PreflightValidation, status *ootov1alpha1.CRStatus,
+		verificationStatus string, message string) error
+	PreflightSetVerificationStage(ctx context.Context, preflight *ootov1alpha1.PreflightValidation,
+		status *ootov1alpha1.CRStatus, stage string) error
```
qbarrand, line 26: Should we have methods updating `Module` and `Preflight` in two distinct interfaces? The `Module` reconciler does not need to update `Preflight` statuses and vice versa.

yevgeny-shnaidman: We can split it into 2 public interfaces, but then it means also initializing to 2 interfaces objects in the main and passing them accordingly to preflight package and to module reconciuler. Keep in mind that in that case we will probably need to define ( in the future) a third inteface, for hub and spoke.  I prefer to keep them all in single interfaces, but i won't object to strongly to splitting them into multiple public interfaces

qbarrand: I don't mind initializing several objects in `main()` and I think it is generally more convenient to have smaller interfaces.

yevgeny-shnaidman: fixed

qbarrand (approved):

ybettan (approved):

PR #132, opened 2022-08-03
Adding Prow jobs to the repo.

ybettan:
Those scripts are expected to be executed by Prow running in the
kubernetes organization.

### DIFF
--- a/ci/prow/lint
+++ b/ci/prow/lint
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+set -e
+
+make lint
--- a/ci/prow/unit_tests
+++ b/ci/prow/unit_tests
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+set -e
+
+make unit-test


ybettan:
Here is the matching PR for configuring Prow - https://github.com/kubernetes/test-infra/pull/27012

ci/prow/lint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
qbarrand, line 1: I would not add the `.sh` extension to those files. Apart from that, LGTM

ybettan: Why not?

qbarrand: The practice of no extension is for all executables, not just shell scripts. See the [Google style guide](https://google.github.io/styleguide/shellguide.html#file-extensions).

ybettan: Have commented on the Prow PR - https://github.com/kubernetes/test-infra/pull/27012

ybettan: Fixed.

qbarrand (approved):

PR #133, opened 2022-08-04
Expand registry package to check existance of a

yevgeny-shnaidman:
file in a v1.Layer ( layer of container image)

This PR includes:
1) adding VerifyModuleExistsForKernel function
2) removing ExtractToolkitRelease funtion: no longer used
   by the preflight code, replaced by the VerifyModuleExistsForKernel
3) add unit-test
4) add test infrastructure to create a dummy v1.Layer

### DIFF
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"path/filepath"
 	"runtime"
 	"strings"
 
@@ -21,7 +22,7 @@ import (
 )
 
 const (
-	driverToolkitJSONFile = "etc/driver-toolkit-release.json"
+	modulesLocationPath = "lib/modules"
 )
 
 type DriverToolkitEntry struct {
@@ -40,7 +41,7 @@ type RepoPullConfig struct {
 
 type Registry interface {
 	ImageExists(ctx context.Context, image string, po ootov1alpha1.PullOptions, registryAuthGetter auth.RegistryAuthGetter) (bool, error)
-	ExtractToolkitRelease(v1.Layer) (*DriverToolkitEntry, error)
+	VerifyModuleExists(layer v1.Layer, pathPrefix, kernelVersion, moduleFileName string) bool
 	GetLayersDigests(ctx context.Context, image string, registryAuthGetter auth.RegistryAuthGetter) ([]string, *RepoPullConfig, error)
 	GetLayerByDigest(digest string, pullConfig *RepoPullConfig) (v1.Layer, error)
 }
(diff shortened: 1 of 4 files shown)


internal/registry/registry.go
```diff
@@ -89,29 +89,11 @@ func (r *registry) GetLayerByDigest(digest string, pullConfig *RepoPullConfig) (
...
-	}
-
-	dtk.OSVersion, found, err = unstructured.NestedString(obj.Object, "RHEL_VERSION")
-	if !found || err != nil {
-		return nil, fmt.Errorf("failed to get RHEL_VERSION from %s, found %t: %w", driverToolkitJSONFile, found, err)
-	}
-	return dtk, nil
+func (r *registry) VerifyModuleExists(layer v1.Layer, pathPrefix, kernelVersion, moduleFileName string) bool {
+	list := []string{pathPrefix, modulesLocationPath, kernelVersion, moduleFileName}
+	fullPath := strings.Join(list, "/")
```
qbarrand, line 94:
```suggestion
	fullPath := filepath.Join(pathPrefix, modulesLocationPath, kernelVersion, moduleFileName)
```

internal/registry/registry_test_helper.go
```diff
@@ -0,0 +1,66 @@
+package registry
```
qbarrand, line 1: Should this file be named `registry_helper_test.go` so it's skipped when compiling the main binary?

internal/registry/registry_test.go
```diff
@@ -335,8 +335,38 @@ var _ = Describe("GetLayersDigests", func() {
...
+		fileName := "/etc/fileName"
+		layer, err := prepareLayer(fileName, []byte("some data"))
+		Expect(err).ToNot(HaveOccurred())
+
+		res := reg.VerifyModuleExists(layer, "", "somekernel", "module_name.ko")
+		Expect(res).To(BeFalse())
+	})
+
+	It("file is present", func() {
+		fileName := "/opt/lib/modules/somekernel/module_name.ko"
```
qbarrand, line 363:
```suggestion
		const fileName = "/opt/lib/modules/somekernel/module_name.ko"
```

internal/registry/registry_test.go
```diff
@@ -335,8 +335,38 @@ var _ = Describe("GetLayersDigests", func() {
...
+		ctrl = gomock.NewController(GinkgoT())
+		reg = NewRegistry()
+	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
+	It("file is not present", func() {
+		fileName := "/etc/fileName"
```
qbarrand, line 354:
```suggestion
		const fileName = "/etc/fileName"
```

internal/registry/registry_test.go
```diff
@@ -335,8 +335,38 @@ var _ = Describe("GetLayersDigests", func() {
...
+	)
+
+	BeforeEach(func() {
+		ctrl = gomock.NewController(GinkgoT())
+		reg = NewRegistry()
+	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
```
qbarrand, line 351:
```suggestion
	reg := NewRegistry()
```

internal/registry/registry.go
yevgeny-shnaidman: fixed

internal/registry/registry_test_helper.go
yevgeny-shnaidman: it will be compiled but no linked, but i will rename it anyway

internal/registry/registry_test.go
yevgeny-shnaidman: fixed

internal/registry/registry_test.go
yevgeny-shnaidman: fixed

internal/registry/registry_test.go
yevgeny-shnaidman: fixed

qbarrand (approved):

ybettan:
Can you please shorten the commit title? (It is messed up because it is too long).

internal/registry/registry_helper_test.go
```diff
@@ -0,0 +1,66 @@
+package registry
```
ybettan, line 1: Why don't we put that code as part of `internal/registry/registry_test.go`?

internal/registry/registry.go
```diff
@@ -89,29 +90,10 @@ func (r *registry) GetLayerByDigest(digest string, pullConfig *RepoPullConfig) (
...
-	}
-
-	dtk.OSVersion, found, err = unstructured.NestedString(obj.Object, "RHEL_VERSION")
-	if !found || err != nil {
-		return nil, fmt.Errorf("failed to get RHEL_VERSION from %s, found %t: %w", driverToolkitJSONFile, found, err)
-	}
-	return dtk, nil
+func (r *registry) VerifyModuleExists(layer v1.Layer, pathPrefix, kernelVersion, moduleFileName string) bool {
+	fullPath := filepath.Join(pathPrefix, modulesLocationPath, kernelVersion, moduleFileName)
+	_, err := r.getHeaderStreamFromLayer(layer, fullPath)
```
ybettan, line 95: Are we going to make use of the io.Reader` in future code?

internal/registry/registry.go
```diff
@@ -89,29 +90,10 @@ func (r *registry) GetLayerByDigest(digest string, pullConfig *RepoPullConfig) (
 	return crane.PullLayer(pullConfig.repo+"@"+digest, pullConfig.authOptions...)
 }
 
-func (r *registry) ExtractToolkitRelease(layer v1.Layer) (*DriverToolkitEntry, error) {
```
ybettan, line 92: How are you going to check that the correct DTK was used in preflight without it?

yevgeny-shnaidman: This is upstream code, we are not going to need it, since we moved from using json file, to check the module path

internal/registry/registry.go
yevgeny-shnaidman: We might, in case we will need to actually access the file data

internal/registry/registry_helper_test.go
yevgeny-shnaidman: since it is quite a few implementation details for mocking the layering, it was better to move it to a helper file, and leave it separate from the test code itself

qbarrand (approved):

PR #134, opened 2022-08-04
RFC: add support for a job to sign kernel modules for secureboot

chr15p:
This is a request for comments to socialise my approach and find out if I've missed any issues. 

This PR adds the ability for kmmo to launch a second type of job alongside the kaniko build job. This second job signs kernel modules by running a binary, whose source is available [here](https://github.com/chr15p/kmod_signer) 

The design is to generalise the existing code taht runs the build by having Reconcile() loop through an array of multiple instances of a manager object (that knows how to start, monitor etc a k8s job). Then have multiple job objects that define the specific job and are run by the manager. 

To do this the PR makes the following changes:
- Add a Sign{} structure to the CRD sitting alongside Build{}, inside KernelMapping{}   
- Rename the "internal/build/"  dir to "internal/jobmanager/" because it does more than just build
- Change the buildAPI field of ModuleReconciler{} to be an array of build.Manager objects, one for each job to be run (so 2)
- Change the reconcile loop to loop through ModuleReconciler.buildAPI[] and run each job in turn (buildAPI is an array so the builder always gets run before signer)
- Make the job.manager class (internal/jobmanager/job/manager.go)  generic so it can run any job that is passed to it on instantiation
- change the Maker interface and maker class to be the Job interface and builder class to make the naming clearer, and then added several methods that implement functionality removed from the manager because it was specific to the type of job
- add a signer class that implements the Job interface and actually defines the job itself (this is the bit I actually set out to do!)

Known issues:
- It uses the old CRD
- Tests in general. I have broken them badly and need to fix them!
- It Works For Me but may still have some subtle bugs
- yes, I should squash my commits.


internal/jobmanager/job/manager.go
```diff
@@ -16,16 +16,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-var errNoMatchingBuild = errors.New("no matching build")
+var errNoMatchingJob = errors.New("no matching job")
 
 type jobManager struct {
```
qbarrand, line 21: I would clearly separate the build and signing management features into two distinct interfaces. I get that the current kaniko job manager and the signing manager are doing similar tasks right now, but:
- upstream, we might move to Shipwright in the future, we means creating a new resource instead of a `Job`;
- downstream, we will work with either Shipwright or BuildConfig - no `Job` either.

Currently, the `Manager` interface in `build` abstracts any build implementation detail, which is convenient when we want to introduce a new build manager - we just provide it as is to the reconciler. The PR as is transforms it into a `Job` manager, which makes sense right now upstream, but might make things more challenging as we add more build systems.

controllers/module_reconciler.go
```diff
@@ -57,7 +57,7 @@ type ModuleReconciler struct {
 
 func NewModuleReconciler(
 	client client.Client,
-	buildAPI build.Manager,
+	buildAPI []build.Manager,
```
qbarrand, line 60: As discussed in another comment, I would have a separate interface for the signing feature.

internal/jobmanager/job/manager.go
```diff
@@ -76,7 +84,16 @@ func (jbm *jobManager) Sync(ctx context.Context, mod ootov1alpha1.Module, m ooto
...
+
+	containerimage, err := jbm.maker.GetOutputImage(mod, &m)
+        if err != nil {
+                return build.Result{}, err
+        }
+
+	pulloptions := jbm.maker.PullOptions(m)
+
+	logger.Info("try to pull image", "img", containerimage, "for job", jbm.GetName() )
+	imageAvailable, err := jbm.registry.ImageExists(ctx, containerimage, pulloptions, registryAuthGetter)
```
qbarrand, line 96: Why are we checking this here?

PR #136, opened 2022-08-05
Add CRD and e2e step test

erusso7:
Add e2e test covering the device plugin management.

Added new labels to distinguish the two different
daemonset types for filtering later on the e2e test.

MGMT-10538

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -133,6 +133,13 @@ jobs:
           done
         timeout-minutes: 1
 
+      - name: Check that the daemon-set for device plugin is running
+        run: |
+          until kubectl get ds  -l 'kmm.node.kubernetes.io/module.name=ooto-ci,kmm.node.kubernetes.io/role=device-plugin' -ojson | jq -e '.items[] | select(.status.numberReady == 1)'; do 
+            sleep 3
+          done
+        timeout-minutes: 1
+
       - name: Remove the Module
         run: kubectl delete -f module-ooto-ci.yaml
 
@@ -267,7 +274,7 @@ jobs:
       - name: Verify that the DaemonSet gets garbage collected
         run: |
           # Cannot use kubectl wait because it will fail if there is no initial match
-          until [ $(kubectl get daemonsets.apps -l kmm.node.kubernetes.io/module.name=ooto-ci-b -o go-template='{{ len .items }}') -eq 0 ]; do
+          until [ $(kubectl get daemonsets.apps -l 'kmm.node.kubernetes.io/module.name=ooto-ci-b, kmm.node.kubernetes.io/role!=device-plugin' -o go-template='{{ len .items }}') -eq 0 ]; do
             sleep 3
           done
         timeout-minutes: 1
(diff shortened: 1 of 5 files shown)


ci/module-ooto-ci.template.yaml
```diff
@@ -12,5 +12,14 @@ spec:
...
           containerImage: ooto-kmod:local
+  devicePlugin:
+    container:
+      image: squat/generic-device-plugin
+      args:
+        - --device
+        - '{"name": "simple-kmod", "groups": [{"paths": [{"path": "/proc/simple-procfs-kmod"}]}]}'
+      volumeMounts:
+        - name: proc
+          mountPath: /proc
```
qbarrand, line 23:
```suggestion
          mountPath: /proc
    volumes:
      - name: proc
        hostPath:
          path: /proc
```

.github/workflows/e2e.yaml
```diff
@@ -133,6 +133,14 @@ jobs:
           done
         timeout-minutes: 1
 
+      - name: Check that the daemon-set for device plugin is running
+        run: |
+          DS_NAME=$(kubectl get ds -l oot.node.kubernetes.io/module.name=ooto-ci -o name)
+          until kubectl get -o json $DS_NAME  | jq -e '.status.numberReady | contains(1)'; do
```
qbarrand, line 139: I think we can have it in one command only:
```suggestion
          until kubectl get ds -l oot.node.kubernetes.io/module.name=ooto-ci,oot.node.kubernetes.io/role=device-plugin -o json | jq -e '.items[] | select(.status.numberReady == 1)''; do 
```

qbarrand:
cc @yevgeny-shnaidman @ybettan - FYI we now have a `oot.node.kubernetes.io/role` label on DaemonSets.

internal/daemonset/daemonset.go
```diff
@@ -105,6 +105,7 @@ func (dc *daemonSetGenerator) SetDriverContainerAsDesired(ctx context.Context, d
 	standardLabels := map[string]string{
 		constants.ModuleNameLabel: mod.Name,
 		dc.kernelLabel:            kernelVersion,
+		constants.NodeRole:        "module-loader",
```
ybettan, line 108: 1. Why is it called `NodeRole` if the label applies to a DS (and not to a node)?
2. Can we use consts instead of `module-loader`?
3. Can we standerize the names? `module-loader` and `device-plugin-loader` or `module` and `device-plugin`?

erusso7: @qbarrand ?

.github/workflows/e2e.yaml
erusso7: The selector should be like the new label `kmm.node.kubernetes.io/role` instead of `oot.node.kubernetes.io`, right?

internal/daemonset/daemonset.go
qbarrand: 1. Indeed, not sure `NodeRole` is appropriate as a constant name - how about `DaemonSetRole`?
2. If it is used only once, I do not see the benefit of using a constant.
3. the first container loads modules, so `module-loader` sounds appropriate to me - see also the PR from @yevgeny-shnaidman https://github.com/qbarrand/oot-operator/pull/148. The device plugin, on the other hand, does not really load anything.

ybettan: Fair enough.

ybettan (approved):

PR #141, opened 2022-08-09
Add toleration key for new label node-role.kubernetes.io/control-plane

enriquebelarte:
From Kubernetes [v1.20](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cluster-lifecycle/kubeadm/2067-rename-master-label-taint/README.md) label "node-role.kubernetes.io/master" is deprecated and "node-role.kubernetes.io/control-plane" will be  used instead.
Adding new toleration key to controller to accept both until the definitive removal of the old label.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -24,8 +24,16 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
-      nodeSelector:
-        node-role.kubernetes.io/master: ''
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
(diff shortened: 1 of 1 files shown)


config/manager/manager.yaml
```diff
@@ -65,3 +65,7 @@ spec:
         - key: node-role.kubernetes.io/master
           operator: Equal
           effect: NoSchedule
+        - key: node-role.kubernetes.io/control-plane
+          operator: Equal
+          effect: NoSchedule
+
```
qbarrand, line 71: Besides that trailing line, lgtm.
```suggestion
```

qbarrand:
Do OpenShift masters also have that label?

PR #142, opened 2022-08-18
Run pods with the system-node-critical priority

qbarrand:
Use the highest standard priority class.

/cc @fabiendupont @ybettan @yevgeny-shnaidman


fabiendupont (approved):
/lgtm

PR #149, opened 2022-08-22
Add preflight package

yevgeny-shnaidman:
Preflight package implements should implement
all preflight functionality per one Module.
Current commit will only implement image verification:
image presence and correct kernel version

MGMT-11438

### DIFF
--- a/internal/module/kernelmapper.go
+++ b/internal/module/kernelmapper.go
@@ -19,14 +19,11 @@ const (
 )
 
 type NodeOSConfig struct {
-	kernelFullVersion  string `subst:"KERNEL_FULL_VERSION"`
-	kernelVersionMMP   string `subst:"KERNEL_XYZ"`
-	kernelVersionMajor string `subst:"KERNEL_X"`
-	kernelVersionMinor string `subst:"KERNEL_Y"`
-	kernelVersionPatch string `subst:"KERNEL_Z"`
-	kernelRHELVersion  string `subst:"KERNEL_RHEL_VERSION"`
-	kernelRHELRelease  string `subst:"KERNEL_RHEL_RELEASE"`
-	kernelRHELArch     string `subst:"KERNEL_RHEL_ARCH"`
+	KernelFullVersion  string `subst:"KERNEL_FULL_VERSION"`
+	KernelVersionMMP   string `subst:"KERNEL_XYZ"`
+	KernelVersionMajor string `subst:"KERNEL_X"`
+	KernelVersionMinor string `subst:"KERNEL_Y"`
+	KernelVersionPatch string `subst:"KERNEL_Z"`
 }
 
 //go:generate mockgen -source=kernelmapper.go -package=module -destination=mock_kernelmapper.go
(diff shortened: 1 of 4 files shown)


internal/module/kernelmapper.go
```diff
@@ -19,14 +19,14 @@ const (
...
-	kernelRHELRelease  string `subst:"KERNEL_RHEL_RELEASE"`
-	kernelRHELArch     string `subst:"KERNEL_RHEL_ARCH"`
+	KernelFullVersion  string `subst:"KERNEL_FULL_VERSION"`
+	KernelVersionMMP   string `subst:"KERNEL_XYZ"`
+	KernelVersionMajor string `subst:"KERNEL_X"`
+	KernelVersionMinor string `subst:"KERNEL_Y"`
+	KernelVersionPatch string `subst:"KERNEL_Z"`
+	KernelRHELVersion  string `subst:"KERNEL_RHEL_VERSION"`
+	KernelRHELRelease  string `subst:"KERNEL_RHEL_RELEASE"`
+	KernelRHELArch     string `subst:"KERNEL_RHEL_ARCH"`
```
qbarrand, line 29: How about we drop those fields and keep them for downstream only? We are about to push this code into Kubernetes.

internal/preflight/preflight.go
```diff
@@ -0,0 +1,99 @@
...
+	osConfig := module.NodeOSConfig{KernelFullVersion: kernelVersion}
+	mapping, err = p.kernelAPI.PrepareKernelMapping(mapping, &osConfig)
+	if err != nil {
+		return false, fmt.Sprintf("Failed to substitute template in kernel mapping in the module %s for kernel version %s", mod.Name, kernelVersion)
+	}
+
+	return p.verifyImage(ctx, mapping, mod, kernelVersion)
+}
+
+func (p *preflight) verifyImage(ctx context.Context, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string) {
```
qbarrand, line 62: Could this just return an error?

internal/preflight/preflight.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+}
+
+type preflight struct {
+	client      client.Client
+	registryAPI registry.Registry
+	kernelAPI   module.KernelMapper
+}
+
+func (p *preflight) PreflightUpgradeCheck(ctx context.Context, mod *kmmv1beta1.Module, kernelVersion string) (bool, string) {
```
qbarrand, line 47: Could this just return an error?

internal/preflight/preflight.go
yevgeny-shnaidman: error won't help us. This should return whether the image is verified, and also the message that describes the reason for verification or not verification

internal/preflight/preflight.go
yevgeny-shnaidman: same comment as above. error wont' tell us much, we need the status and the reason message for the status

internal/preflight/preflight.go
qbarrand: We could do:
- image verified: depends on whether `err` is nil or not
- reason: the error string
WDYT?

internal/module/kernelmapper.go
yevgeny-shnaidman: fixed

internal/preflight/preflight.go
yevgeny-shnaidman: I would prefer not to: error means that something went wrong with the flow, while in preflight case there is nothing wrong with the flow, the function  just returns the status of verification.

qbarrand (approved):

PR #150, opened 2022-08-23
Add toleration key for new label node-role.kubernetes.io/control-plane

enriquebelarte:
From Kubernetes [v1.20](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cluster-lifecycle/kubeadm/2067-rename-master-label-taint/README.md) label "node-role.kubernetes.io/master" is deprecated and "node-role.kubernetes.io/control-plane" will be  used instead.
Adding new toleration key to controller to accept both until the definitive removal of the old label.

### DIFF
--- a/ci/install-ci/kustomization.yaml
+++ b/ci/install-ci/kustomization.yaml
@@ -9,16 +9,6 @@ images:
     newName: ooto
     newTag: local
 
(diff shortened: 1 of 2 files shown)


qbarrand (approved):
Thank you!

PR #152, opened 2022-08-24
Load firmware from the module loader

fabiendupont:
It is common for kernel module to come with firmware(s). On immutable operating systems like CoreOS, the standard firmware path `/lib/firmware` is read-only. So, we need to let the kernel know where to lookup firmware file(s).

This change adds the `FirmwarePath` attribute to the `ModprobeSpec` definition and modify the `MakeLoadCommand` and `MakeUnloadCommand` methods to leverage it.

It also adds a `Volume` and a `VolumeMount` to expose the `/var/lib/firmware` host directory (created if it doesn't exist) inside the pod. The `/var/lib/firmware` is [the recommended path for firmwares in OpenShift with RHCOS nodes](https://access.redhat.com/documentation/fr-fr/openshift_container_platform/4.10/html-single/post-installation_configuration/index#rhcos-load-firmware-blobs_post-install-machine-configuration-tasks).

The `MakeLoadCommand` adds a step to copy the `FirmwarePath` to `/var/lib/firmware` to the `modprobe` command. The `MakeUnloadCommand` removes the firmware files previously copied after unloading the module. The commands use `/usr/bin -c` to allow shell control like `&&`.

Signed-off-by: Fabien Dupont <fdupont@redhat.com>


internal/daemonset/daemonset.go
```diff
@@ -340,42 +364,73 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
...
+	loadCommand := "modprobe"
 
 	if ra := spec.RawArgs; ra != nil && len(ra.Load) > 0 {
-		return append(loadCommand, ra.Load...)
+		loadCommand = fmt.Sprintf("%s %s", loadCommand, strings.Join(ra.Load[:], " "))
+		return append(loadCommandShell, loadCommand)
+	}
+
+	if fw := spec.FirmwarePath; fw != "" {
+		loadCommand = fmt.Sprintf("cp -r %s %s && %s", fw, nodeVarLibFirmwarePath, loadCommand)
```
qbarrand, line 380: How about doing this in an InitContainer instead?

fabiendupont: It wouldn't be symmetric with `preStop` hook, as there's not "`exitContainer`". The pod spec remains fairly simple with a single container that does everything.

qbarrand: The `exitContainer` could be a regular `container` sleeping with a `preStop` hook. We would not have to move from an `exec`-type command to a shell one.
@yevgeny-shnaidman WDYT?

fabiendupont: A sleeping container could generate a race condition with the main container, as you wouldn't know which preStop gets executed first. If the module needs the firmware during is unload phase, it may not find it if the `exitContainer`'s preStop runs first.

PR #158, opened 2022-08-25
Add preflight controller implementation

yevgeny-shnaidman:
This PR implements the preflight controller logic
and filtering

1) add filter function for getting preflights
   on modules change
2) add controller implementation
3) unit-test


controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+
+// ClusterPreflightReconciler reconciles a PreflightValidation object
+type PreflightValidationReconciler struct {
+	client        client.Client
+	filter        *filter.Filter
+	statusUpdater statusupdater.PreflightStatusUpdater
+	preflight     preflight.PreflightAPI
+}
+
+func NewPreflightReconciler(
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("preflightvalidation").
+		For(&kmmv1beta1.PreflightValidation{}).
+		Watches(
+			&source.Kind{Type: &kmmv1beta1.Module{}},
+			handler.EnqueueRequestsFromMapFunc(r.filter.FindPreflightsForModule),
+			builder.WithPredicates(r.filter.PreflightReconcilerModulePredicate()),
+		).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: 1,
```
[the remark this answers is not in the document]
yevgeny-shnaidman: yes, but i like to explicitly state it, always make it clear for the user, and illuminates any charges of race conditions :)

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+// Reconcile Reconiliation entry point
+func (r *PreflightValidationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
+	reconRes := ctrl.Result{}
+
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("Start PreflightValidation Reconciliation")
+
+	pv := kmmv1beta1.PreflightValidation{}
+	err := r.client.Get(ctx, types.NamespacedName{Namespace: req.Namespace, Name: req.Name}, &pv)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("PreflightValidation reconcile success. Reconciliation object not found, probably deleted. Not reconciling")
+			return reconRes, nil
+		} else {
+			log.Error(err, "preflight validation reconcile failed to find object")
+			return reconRes, err
+		}
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+		log.Error(err, "runPreflightValidation failed")
+		return reconRes, err
+	}
+
+	if reconCompleted {
+		log.Info("PreflightValidation reconciliation success")
+		return ctrl.Result{}, nil
+	}
+	log.Info("PreflightValidation reconciliation requeue")
+	return ctrl.Result{RequeueAfter: time.Second * reconcileRequeueInSeconds}, nil
```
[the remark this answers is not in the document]
yevgeny-shnaidman: we want to reschedule the reconciliation, so that user might fix the validation. We cannot count on Module update, since the correction might include only pre-building an image

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func (r *PreflightValidationReconciler) runPreflightValidation(ctx context.Context, pv *kmmv1beta1.PreflightValidation) (bool, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	modulesToCheck, err := r.getModulesToCheck(ctx, pv)
+	if err != nil {
+		return false, fmt.Errorf("failed to get list of modules to check for preflight:%w", err)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+
+func (r *PreflightValidationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("preflightvalidation").
+		For(&kmmv1beta1.PreflightValidation{}).
+		Watches(
+			&source.Kind{Type: &kmmv1beta1.Module{}},
+			handler.EnqueueRequestsFromMapFunc(r.filter.FindPreflightsForModule),
+			builder.WithPredicates(r.filter.PreflightReconcilerModulePredicate()),
+		).
```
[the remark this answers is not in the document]
yevgeny-shnaidman: This is a compromise: we have not yet reached decision on when an how we are going to stop reconciliation, we are also watching the changes in the modules for now. Maybe we might change it for only newly created ones. By the end of the epic, this topic will be finilized

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
+/*
```
ybettan, line 1: All other controllers in `controllers/` are named `<CR name>_reconciler.go`. Should we rename?

controllers/preflightvalidation_controller_test.go
```diff
@@ -0,0 +1,315 @@
+package controllers
```
ybettan, line 1: Same as https://github.com/qbarrand/oot-operator/pull/158/files#r956685251

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+	reconRes := ctrl.Result{}
+
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("Start PreflightValidation Reconciliation")
+
+	pv := kmmv1beta1.PreflightValidation{}
+	err := r.client.Get(ctx, types.NamespacedName{Namespace: req.Namespace, Name: req.Name}, &pv)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("PreflightValidation reconcile success. Reconciliation object not found, probably deleted. Not reconciling")
```
ybettan, line 93: So the trigger for entering the reconciliation loop in that case is the Preflight CR deletion?!

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+			return reconRes, nil
+		} else {
+			log.Error(err, "preflight validation reconcile failed to find object")
+			return reconRes, err
```
ybettan, line 97:
```suggestion
			return ctrl.Result{}, err
```
It is the same but clearer IMO. You also do it later on this file anyway.

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+
+	reconCompleted, err := r.runPreflightValidation(ctx, &pv)
+	if err != nil {
+		log.Error(err, "runPreflightValidation failed")
+		return reconRes, err
```
ybettan, line 104:
```suggestion
			return ctrl.Result{}, err
```
It is the same but clearer IMO. You also do it later on this file anyway.

controllers/preflightvalidation_controller.go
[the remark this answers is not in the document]
ybettan: What is the default requeue time if we only use `ctrl.Result{Requeue: true}`?

controllers/preflightvalidation_controller_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+		res, err := pr.Reconcile(ctx, req)
+
+		Expect(err).To(BeNil())
+		Expect(res).To(Equal(reconcile.Result{RequeueAfter: time.Second * reconcileRequeueInSeconds}))
+	})
+
+})
+
+var _ = Describe("GetModulesCheck", func() {
```
ybettan, line 170:
```suggestion
var _ = Describe("getModulesToCheck", func() {
```

I suggest we keep the `Describe` name identical with the function/method name it is testing.

controllers/preflightvalidation_controller_test.go
```diff
@@ -0,0 +1,315 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	preflightName = "test-preflight"
+)
+
+var _ = Describe("PreflightValidationReconciler", func() {
```
ybettan, line 26:
```suggestion
var _ = Describe("Reconcile", func() {
```

I suggest we keep the `Describe` name identical with the function/method name it is testing.

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+	return r.checkPreflightCompletion(ctx, pv.Name, pv.Namespace)
+}
+
+func (r *PreflightValidationReconciler) getModulesToCheck(ctx context.Context, pv *kmmv1beta1.PreflightValidation) ([]kmmv1beta1.Module, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	modulesList := kmmv1beta1.ModuleList{}
+	err := r.client.List(ctx, &modulesList)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get list of all Modules, %w", err)
```
ybettan, line 142:
```suggestion
		return nil, fmt.Errorf("failed to get list of all Modules: %w", err)
```

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	moduleStatus := r.getModuleStatus(pv, moduleName)
+	if moduleStatus == nil {
+		log.Info(utils.WarnString("failed to find the status of Module CR in preflight"), "module", moduleName)
+		return
+	}
+	err := r.statusUpdater.PreflightSetVerificationStatus(ctx, pv, moduleStatus, verificationStatus, message)
+	if err != nil {
+		log.Info(utils.WarnString("failed to update the status of Module CR in preflight"), "module", moduleName, "error", err)
```
ybettan, line 192: Why don't we use `log.Error` here?

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+
+	return true, nil
+}
+
+func (r *PreflightValidationReconciler) getModuleStatus(pv *kmmv1beta1.PreflightValidation, moduleName string) *kmmv1beta1.CRStatus {
+	for i := range pv.Status.CRStatuses {
+		if pv.Status.CRStatuses[i].Name == moduleName {
+			return &pv.Status.CRStatuses[i]
+		}
+	}
```
ybettan, line 237: How about using a a map instead of a slice as @qbarrand suggested?

main.go
```diff
@@ -158,6 +159,8 @@ func main() {
 	daemonAPI := daemonset.NewCreator(client, kernelLabel, scheme)
 	kernelAPI := module.NewKernelMapper()
 	moduleStatusUpdaterAPI := statusupdater.NewModuleStatusUpdater(client, daemonAPI, metricsAPI)
+	preflightStatusUpdaterAPI := statusupdater.NewPreflightStatusUpdater(client)
+	preflightAPI := preflight.NewPreflightAPI(client, registryAPI, kernelAPI)
```
ybettan, line 163: Why do we have 2 different interfaces for Preflight?

controllers/preflightvalidation_controller.go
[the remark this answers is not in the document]
yevgeny-shnaidman: we want to control the timeout. In our case we don't want the re queuing to happen quickly

controllers/preflightvalidation_controller.go
yevgeny-shnaidman: reconciliation loop is watching the Preflight CR, so any change of CR status will trigger reconciliation. I will check if we might add a filter for deletion events, also it won't help in that case, since we still need to check the case where resource was already deleted

controllers/preflightvalidation_controller.go
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller.go
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller_test.go
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller_test.go
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller.go
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller.go
yevgeny-shnaidman: Since it does not stop the flow the same way as an actual error does. In this case we continue with the loop and we continue with the processing

main.go
yevgeny-shnaidman: we have interface for updating the statuses of the KMM objects (module, preflight, maybe later for hub/spoke), and we have interface for preflight processing

internal/filter/filter.go
```diff
@@ -46,6 +46,10 @@ func (f *Filter) ModuleReconcilerNodePredicate(kernelLabel string) predicate.Pre
 	)
 }
 
+func (f *Filter) PreflightReconcilerModulePredicate() predicate.Predicate {
+	return predicate.GenerationChangedPredicate{}
```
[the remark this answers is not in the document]
ybettan: How about making it a function instead of a method? It doesn't depend on the underlying object.

internal/filter/filter.go
```diff
@@ -107,6 +111,28 @@ func (f *Filter) FindModulesForNode(node client.Object) []reconcile.Request {
 	return reqs
 }
 
+func (f *Filter) FindPreflightsForModule(mod client.Object) []reconcile.Request {
```
ybettan, line 114: I am not sure I am getting this method:
1. Should we rename that method? Not sure the name fit what the method does.
2. Why are we using a `client.Object` and not a KMM `Module`?
3. Why do we need to set the logger with the module name if we get all the Prefligh objects blindly anyway?

controllers/preflightvalidation_controller.go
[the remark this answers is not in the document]
ybettan: I agree, just wondering what is the default time.

main.go
ybettan: Why don't we use a single interface? or a single interface that embed both smaller interfaces if you prefer the separation?
I just wonder why the Reconciler need to different interfaces for manipulating Preflight.

yevgeny-shnaidman: that's was a discussion that we had when creating/updating statusupdater package. we use this package as an updater for all the object of KMM, and it provides different interfaces for different objects

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func (r *PreflightValidationReconciler) getModulesValidationStatusesMap(pv *kmmv1beta1.PreflightValidation) map[string]kmmv1beta1.CRStatus {
+	statusMap := make(map[string]kmmv1beta1.CRStatus, len(pv.Status.CRStatuses))
+	for _, status := range pv.Status.CRStatuses {
+		statusMap[status.Name] = status
+	}
+
+	return statusMap
+}
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller.go
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller_test.go
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller.go
yevgeny-shnaidman: fixed

internal/filter/filter.go
yevgeny-shnaidman: if you have a better name, i will use it. 
we are using the client.Object because this is the interface that is provided by the k8s for mapping functions of the Watch
i am setting it to module name, since this function is activated on module change.

internal/filter/filter.go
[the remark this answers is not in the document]
yevgeny-shnaidman: @qbarrand i don't this we should skip modules deletion: it may be a valid fix from users point of view. @ybettan , fixed

controllers/preflightvalidation_controller.go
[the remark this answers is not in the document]
qbarrand: Why not use the default backoff scheme?

internal/filter/filter.go
```diff
@@ -107,6 +107,28 @@ func (f *Filter) FindModulesForNode(node client.Object) []reconcile.Request {
...
+	logger.Info("Listing all preflights")
+	preflights := kmmv1beta1.PreflightValidationList{}
+	if err := f.client.List(context.Background(), &preflights); err != nil {
+		logger.Error(err, "could not list preflights")
+		return reqs
+	}
+
+	for _, preflight := range preflights.Items {
+		// skip the preflight being deleted
+		if preflight.GetDeletionTimestamp() != nil {
```
qbarrand, line 123: I think we already make that check in the reconciler code?

internal/filter/filter.go
qbarrand: How about `EnqueueAllPreflightValidations`?

controllers/preflightvalidation_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+// Reconcile Reconiliation entry point
+func (r *PreflightValidationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("Start PreflightValidation Reconciliation")
+
+	pv := kmmv1beta1.PreflightValidation{}
+	err := r.client.Get(ctx, req.NamespacedName, &pv)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("PreflightValidation reconcile success. Reconciliation object not found, probably deleted. Not reconciling")
```
qbarrand, line 91:
```suggestion
			log.Info("Reconciliation object not found; not reconciling")
```

controllers/preflightvalidation_controller.go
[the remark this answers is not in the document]
yevgeny-shnaidman: i don't see the point to use it. backoff scheme is used in case of an errors, in our case we just want to give a user a specific amount of time to fix something from the beginning. The current timeout value will be for sure changed once we have customer using this feature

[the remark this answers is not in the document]
qbarrand: OK then.

internal/filter/filter.go
yevgeny-shnaidman: works for me, fixed

internal/filter/filter.go
yevgeny-shnaidman: no, we check there the modules' deletion condition, not the preflights'

controllers/preflightvalidation_reconciler.go
yevgeny-shnaidman: fixed

qbarrand (approved):

PR #160, opened 2022-08-29
Trim plus sign in kernel versions

enriquebelarte:
Google Kubernetes Engine uses COS as default OS nodes. COS kernel versions end in '+' sign which is not a valid ending for Kubernetes labelling syntax.

### DIFF
--- a/controllers/node_kernel_reconciler.go
+++ b/controllers/node_kernel_reconciler.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/qbarrand/oot-operator/internal/filter"
 	v1 "k8s.io/api/core/v1"
@@ -37,7 +38,7 @@ func (r *NodeKernelReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return ctrl.Result{}, fmt.Errorf("could not get node: %v", err)
 	}
 
-	kernelVersion := node.Status.NodeInfo.KernelVersion
+	kernelVersion := strings.TrimSuffix(node.Status.NodeInfo.KernelVersion,"+")
 
 	logger.Info(
 		"Patching node label",


controllers/node_kernel_reconciler.go
```diff
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+        "strings"
```
qbarrand: I believe there are too many tabs here.

controllers/node_kernel_reconciler.go
```diff
@@ -37,7 +38,7 @@ func (r *NodeKernelReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, fmt.Errorf("could not get node: %v", err)
 	}
 
-	kernelVersion := node.Status.NodeInfo.KernelVersion
+	kernelVersion := strings.TrimSuffix(node.Status.NodeInfo.KernelVersion,"+")
```
qbarrand, line 41:
```suggestion
	kernelVersion := strings.TrimSuffix(node.Status.NodeInfo.KernelVersion, "+")
```
We also need a unit test for this feature.

qbarrand: Wondering if this behaviour should be optional. @ybettan @yevgeny-shnaidman thoughts?

yevgeny-shnaidman: @qbarrand why do you think it should be optional? In case it is a valid kernel version string, it won't be changed. What use case do you see for keeping the "+" at the end of the kernel version?

controllers/node_kernel_reconciler.go
yevgeny-shnaidman: @qbarrand don't we have a linter check running on each PR?

controllers/node_kernel_reconciler.go
qbarrand: I do not see any, really. I am just thinking that customers might be adding patches that would be required by their kmod. But then those customers would probably be managing their kernel releases like RHEL does.

